PR #40, opened 2022-10-04
fix: able to run the Vue.js demo

Zeng95:
Installed three.js to be able to run the vue.js demo


remcoder:
Hi @Zeng95 thanks for your PR!

I cannot reproduce the issue. Were you using and older version of three.js? I did a fresh `git clone` and the vue demo just worked. What do you think the issue is?

remcoder:
fixes #41

Issue #55, opened 2023-01-13
Implement support for G2 / G3 arcs


RickRyan26:
I'd love to use this package but I need arcs.
Is there a way to use G2/G3/G02/G03 with gcode-preview?

remcoder:
Hi @RickRyan26, G2/G3 arcs are currently not supported. I don't have the time right now to dive into it but I'd love to support this at some point. It has been on my list for a while now. 

I could sure use any kind of help, like a PR, to get started, if you're up for it.  

Otherwise you might look at this other gcode viewer by @Sindarius, which supports G2/G3 arcs and it is overall really good: https://github.com/Sindarius/npm_gcodeviewer

Sindarius:
If you look in my source you can adapt doArc@src/utils.js to add support for g2/g3 in your viewer without too much work I bet.

remcoder:
Thx @Sindarius! That will be helpful :-)

RickRyan26:
I created a fork, tried to add the G2 functionality based on Sindarius' doArc, and thought I had it, but all my arcs come out flat lol was a fun attempt. Thanks for the thoughtful comments above

remcoder:
I'll reopen this issue because it is a valid feature request that should get addressed at some point. Also with the example code and your first steps it would be a shame to let that go to waste.

remcoder:
@RickRyan26 A new version was pushed to npm just now. Let us know if it works for you.

RickRyan26:
@remcoder I scrolled through your changes, awesome work, much more complete than my attempt!

And I just tried your latest version but the following seems to still draw a straight line.
`const gCode = 'G0 X0 Y0 Z0.2\nG2 X42 Y42 R21 E10';`

remcoder:
There seems to be an unrelated bug causing the first extrusion to be skipped sometimes. Could you try adding 1 or 2 G0/G1 extrusions before the G2? 

Otherwise it might be in the code that calculates I and J from the R param, which I didn't test well enough yet. 

I will look at the latter first and report back.

Sindarius:
G0 X0 Y0 Z0.2
G2 X42 Y42 R21 E10
Based on those two lines of gcode the G2 is an incorrect value.  R21 would not fit the math.  The minimum R value can't be 21
Find the midpoint between the two points
x = 21, y =21
Apply the Pythagorean theorem to find our R to point 42,42 from 21,21
sqrt(21^2 + 21^2)
[...]

PR #56, opened 2023-01-15
Implement support for G2/G3 arcs

remcoder:
This PR, in response to https://github.com/remcoder/gcode-preview/issues/55,  aims to implement support for G2 and G3 commands to produce arcs.

Following [Marlin](https://marlinfw.org/docs/gcode/G002-G003.html), The output of this would be to create an approximation of the defined arcs using small (~1mm) straight line segments and pass that on to the Three.js renderer. 

image

[This code in npm_gcode_viewer can be used for inspiration](https://github.com/Sindarius/npm_gcodeviewer/blob/main/src/utils.js).


Sindarius:
I flubbed and probably should have done the work on this here but @RickRyan26 I have a PR setup which fully implements G2/G3 and the I/J params. Tools like ArcWelder are going to utilize both so you have to have both implemented or you can end up with missing points in the gcode which lead to circle artifacts.  There is one issue where the top of the test bench is not fully rendering so feel free to look into that.

Arc benchy
image

Same benchy (you'll see G2/G3)
image

remcoder:
@Sindarius Awesome! Thx a lot for helping out dude!

Sindarius:
Not a problem once he pulls my pr across into his you can grab and test and find that one last issue. My guess is maybe a count is off somewhere but I am not certain. But this will get you close 👍

remcoder:
Super! 
@RickRyan26 I'm already pulling in the changes from both your repo's so don't worry.

remcoder:
@RickRyan26 changes have been pushed to the `feature/arcs` branch for now. Please give it a try. I'll publish a new version on npm soon

PR #82, opened 2023-11-20
add tolerance property to parser to improve performance in preview

raulodev:
When you try to render a spiral type gcode or one with many layers, the preview performance looks slow, my solution reduces the number of layers by adding a property called tolerance so that a layer is created only when it is exceeded


remcoder:
great idea! thx @raulodev  😄

remcoder:
This was just released in [v2.11](https://github.com/remcoder/gcode-preview/releases/tag/v2.11). Thanks again for this fix! ;-)

PR #86, opened 2023-12-15
fix webgl-preview.ts

raulodev:
When  paramater z is 0 in some commands the preview shows this

to fix this change from :
> `if (next.z)`
to
> `if (next.z >= 0)`

Then it shows as it should


remcoder:
hi @raulodev thanks for your fix. Much appreciated!

PR #88, opened 2023-12-18
fix webgl-preview.ts

raulodev:
in this short gcode the travel line not shows

> G0 X0 Y0 Z0
>G1 X0 Y0 Z10 E0

to show line of travel completely vertical add this
`|| next.z != state.z`


remcoder:
@raulodev good catch! 😄

Issue #89, opened 2023-12-18
Render extrusions as tubes


sophiedeziel:
I've been playing a lot with the code and ended up with a prototype that renders extrusions as tubes instead of lines. The idea is to be able to render a more advanced preview that resembles what PrusaSlicer and Cura have. I was wondering what are your plans for future versions of this tool because this is not a minor change and is worth planning carefully. 

Here is a screenshot of what I was able to render:

<img width="1680" alt="Screenshot 2023-12-18 at 5 12 11 PM" src="https://github.com/remcoder/gcode-preview/assets/1037105/015e1373-87a1-44d5-9bc3-50cbe5a9b8b1">

Extrusion paths are tubes and travel moves are still lines.

I was thinking about switching from lines to tubes using a boolean in the preview options. 

However, my prototype is very far from performant and is not accurate enough in my opinion. I'm still learning Three.js and feedback is welcome.

If I submit pull-requests, how would you prefer to proceed? Of course, smaller PRs with a clear scope are preferable. Is there a list of options and calculations that matters to you? Here is what I had in mind:

- Option to switch from lines to tubes
- When the metadata is available, different colors for outer lines, inner lines, and infill
- Accuracy in the dimensions of the tubes for layer height and line width
- Transparency of travel move lines

Let me know what you think. There are a lot of cool things I can add to this project and I'd prefer to contribute here rather than using a fork.

remcoder:
Wow the screenshot looks really good! 

To answer your question first: I'm not dedicating any time to developing on the project myself atm. This is mostly due to time and life priorities. That being said, I've always answered issue's and PR's. And sometimes this sparks a moment where add some improvements myself. 

So I guess if you want to use this code as a base, please go ahead. Maybe if you want to go really fast, a fork of your own might be the best way to go. If you contribute here, on the other hand, I, and others who follow this repo, can assist you here and there. 

I would personally really love if you collaborate on this project, and I'm open to give you access to the repo. Just remember that I cannot make any promises about my own time.

About the tubes: I have several things I'm thinking about:

Rendering extrusion paths with volume can be done in a few ways, including round tubes. Maybe the simplest to start with would be to use square tubes. This results in a minimum of triangles in the mesh, as opposed to round tubes. And some basic things like line width and centering would need to be tackled anyway. 

Another route would be to use the [model used by Slic3R / PrusaSlicer](https://manual.slic3r.org/advanced/flow-math). 
[...]

Issue #85, opened 2023-11-30
Possibility provide custom range for brightness


ivan-hryshko:
I have a problem with displaying black (#000000), gray(#808080) and white(#FFFFFF) they all look the same , 
Is it possible to set range for:
black color for example brightness in the range [0-30]%
for gray [30-70]%
And for white [70-100]%
I think it will make the display of models more realistic
<img width="367" alt="image" src="https://github.com/remcoder/gcode-preview/assets/13918933/dcb846d2-c18c-4c69-8a9f-d0a676903ac9">

remcoder:
Good point. It is a limition of the current implementation.

I'm trying to think of a way that works also for non black/gray/white colors. And I'm reluctant to add more config options but I'm also reluctant to change existing behavior.

Do you have some ideas?

Also to get this moving a PR and some test cases would be really helpful 😄

ivan-hryshko:
@remcoder  i dont have permision to make PR, but i. thing about something like this
 ```
 minBrightness: number = 10;
 maxBrightness: number = 70;`
```

```
    this.maxBrightness = this.validateBrightness(opts.maxBrightness) ?? this.maxBrightness;
    this.minBrightness = this.validateBrightness(opts.maxBrightness) ?? this.minBrightness;
```
 ```
validateBrightness(value: number) {
    if (value === undefined || value === null || typeof value !== 'number') {
      return undefined
    }
    if (value > 100) {
      return 100
    }
    if (value < 0) {
      return 0
    }
    return value
  }
```

 ```
     
        const preparedMaxBrightness = this.maxBrightness / 100
        const preparedMinBrightness = this.minBrightness / 100
        const brightness = preparedMinBrightness + (preparedMaxBrightness * index) / this.layers.length;
```

remcoder:
Thanks for your response 🙂 

It would work but it also feels a little bit ad hoc.

I want the preview to Just Work for most gcode files without much tuning. But you're right that currently the shading doesn't work well with white/gray/black.

So I'm inclined to have a built-in way (so no extra settings) of reducing the brightness range when grey colors are used (low saturation cases).

But we also need to keep in mind that this isn't true lighting so it will never look as good. For instance the more we reduce the brightness range, the harder it is so see detail. 

What are your ideas on this?

FYI, if you want to know how you can send a PF: create a fork of this repo first, commit changes to your forked repo, and then send me a PR.

ivan-hryshko:
[...]

Issue #71, opened 2023-09-29
Vase Mode Previewing


FlatTony:
I am seeing lots of lag and unresponsiveness when loading files that are sliced in vase mode (constant change in Z). The files are less than 15Mb so it should be able to handle it fine from the point density. I am attaching a file that will not render. The bed size is 635,635,635 for reference. If this can be resolved, it is much appreciated. I will be working on it as well and report back with any findings. 

Cheers.

[FlowerPot.zip](https://github.com/remcoder/gcode-preview/files/12771696/FlowerPot.zip)

remcoder:
Hi @FlatTony, ah yes this must be caused by the fact that a new layer is created in memory for each unique z height. Then gcode commands are grouped by layer. This offers some benefits like easy rendering up to a certain layer but it als also has some drawback.. and vase mode especially suffers. 

I think in the end grouping by layer wasn't the best idea but it's going to take rewriting parts of the parser / renderer to really fix this. For now this isn't something I'm going to spend time on myself. 

If you feel like having a go a PR yourself, I'm more than happy to help your out where I can. 

cheers.

remcoder:
fixed in #82

remcoder:
@FlatTony this was fixed in v2.11 which was just released. The FlowerPot loads fine now ;-)

FlatTony:
@remcoder This was the same fix that I came up with as well. Looks good, thanks!

PR #92, opened 2023-12-21
Add a mode to render extrusion as tubes

sophiedeziel:
As discussed in https://github.com/remcoder/gcode-preview/issues/89, here is a first version of rendering extrusions as tubes.

There are some key differences from the prototype as well as some decisions that I've made. I'm open to feedback and I can revert/adapt them:

1. Travel moves are always lines, not tubes. It helps with visibility and performance
2. The `Vector3` defined in `webgl-preview.ts` conflicted with `Three.js`'s class. I renamed it to `GVector3`. I did not put too much thinking in the name. I thought they came from GCode, so they are GVectors. I you prefer a different name or switch to `Three.js`'s vectors, I'll make the change.
3. I changed how Single Layer mode renders colors for lines AND tubes. It takes the pure colors without changing the brightness. A single layer is about the details and they are much easier to see when the color is an exact match to what was chosen.
4. Rendered tubes don't have the gradient as well. I can change that, especially if https://github.com/remcoder/gcode-preview/issues/85 is implemented.
5. As discussed in the original issue, `TubeGeometry` with a `CatmullRomCurve3` is much more performant than single-cylinder segments. It's prettier as well.
6. I've evaluated that a factor of `2` for the number of tube segments per tube length units is the best balance between performance versus render quality. This could be a variable in the future.
7. `MeshLambertMaterial` is the most performant, but switching to one that offers more options in terms may give us more flexibility in the future. The performance gains are not humanly noticeable with the benchy demo as of right now.
8. I set a tube radius of `0.3`, but it's arbitrary. It could be tied to `lineWidth`, have its own new option, or be calculated based on the slicer's line width or layer height, or a combination of those variables. I'm keeping that for a future PR.

Here is a screenshot of the final result!
Screenshot 2023-12-20 at 7 14 08 PM


remcoder:
## Feedback on your points

Totally agree on all points. Both on the current choices and what to hold back for later. 

Naming my own type Vector3 was a mistake. The only worry I have is that is inside the .d.ts file, which makes it an exported type. Renaming it can cause consumers to have build failures. I don't consider it to be high-risk or important enough to prevent this though. So let's stay with GVector3 ;-)

Let's not add a gradient here. It was only a cheap way to mimic lighting.

## Finishing up 
So, if I'm not mistaken this might already be the point were the PR can finished, not? You didn't mention any significant TODO's and I don't have any in terms of functionality either.

I'd prefer to make a separate `addTubeLine` (or something) instead of adding cases inside `addLine`. This also removes the need for `forceLine` for travel lines bc it can just use the old addLine as it were.

There's some things like the lighting that are added to the scene but are only needed for realistic rendering.

I spotted a [typo](https://github.com/remcoder/gcode-preview/pull/92/files#diff-8ec787c979ef6cb88e426e798a75cc41d326888eaca61c5f5691b18c5e65c729R51) btw. 

The only 'big' question is how to expose this new behaviour. I think it should be opt-in, do you agree? And the config param to turn it on should be not too prone to typo's so I'd be more in favour of something like `renderTubes` than `realisticExtrusion` ;-) 

## Testing & what were aiming for
Let's also test with at least:
 - a model that is 2-3x the size as a benchy or even bigger
 - a [vase](https://github.com/remcoder/gcode-preview/issues/71)
 
For now, it is OK for this rendering mode to be unusable for a certain domain of large models. As long as it is usable for a reasonable set of models. 

## Guidelines
As for general guidelines, just as an FYI and since you asked for my vision, here are some things that are in the back of my head:

- the library should Just Work for any gcode file that comes out of a slicer.
- CNC compatibility comes second if the choice needs to be made
- the preview should work with a minimum of mandatory config to be set.
- Tweakability comes after that and we need to be careful not to duplicate THREE's api and pass all the params through. 
- Backwards compatibility is important (I think this is how you enable this slow growth of a small OSS project). Both in terms of API and in terms of behaviour. (I think I broke that principe when I introduced customizabe colors but it's an exception). Performance is important because it can easily become a problem. Both slowdowns and memory usage going through the roof have been issues in the past.
- it should in principle work on mobile without modification, for reasonably sized models. 
 
 I'll add these to a more visible document too.

remcoder:
Oh and did I mention you're doing amazing work?.. You're doing amazing work! 😉  🚀

sophiedeziel:
I've implemented all of your suggestion, I agree with them. Thank you for the feedback!

Also, I tried a vase, a file with multiple parts and a large print. As I expected, it takes much longer to render. But it is still in the useable range. It bothers me that it locks the whole tab's operations, I'd like to work on this next. If you are OK with this, I think it should be a follow up PR. 

## The large vase you linked
This one highlights the need for a dynamic tube radius. It renders nicely, but the layer height is obviously larger. Also keeping this for a follow-up PR.

Screenshot 2023-12-22 at 10 45 04 PM

## Multiple parts

Screenshot 2023-12-22 at 10 49 11 PM

## A large print
This one took 9 seconds in average to render.

Screenshot 2023-12-22 at 10 51 15 PM

remcoder:
merry 🎄! (if you're celebrating)

Works great & looks great! The render times are similar for me. (Changing color is actually very fast comparatively.) 

Nice work @sophiedeziel

PR #101, opened 2024-01-16
Fix: demo UI sidebar not updating

remcoder:
- Implemented drag n drop in de demo code
- on drop: update the sidebar (filename, size, #of layers, thumbnail if present)
- add deprecation notice for drag n drop as a lib feature. 

collateral: 
- since a freshly dropped file isn't animated, it doesn't make much sense to highlight the top layer so it was unset in this case

@sophiedeziel any feedback before I merge this?


sophiedeziel:
I tested those changes and the code LGTM!

sophiedeziel (approved):

PR #102, opened 2024-01-16
Feature: add params to toggle rendering of travel/extrusion

remcoder:
Toggling rendering of travel and extrusion via a param passed to init() was apparently forgotten. Now these are added which allows the full state of the component to specified in 1 statement.

API changes: 

GCodePreviewOptions: 
\+ renderExtrusion: boolean
\+ renderTravel: boolean


sophiedeziel (approved):
I'm wondering at which point we'll want to consider having a structured interface instead of a flat list of params. As long as there are good defaults and minimal required params, there is nothing wrong with a large number of options, but imagine if it looked like this:

```js
extrusion: {
  render: true,
  color: 'hotpink',
  style: 'tubes',
},
travel: {
  render: true,
  transparency: 0.8,
  lineWidth: 1.5
 },
```
There's imaginary options in there, of course. 

Just pitching an idea before they become options to support. This PR is good to go as it is, but it indicates that the list of params may grow

remcoder:
Yeah I know what you mean. I think we can keep that in mind for 3.0. At that point I'd like to drop the init() function as well and just use a constructor.

PR #103, opened 2024-01-16
Fix incorrect segmenting of extrusion lines when line width is set.

### LINKED ISSUE #100: Some travel lines renders as extrusion lines
Hi, I've encounter a bug when setting the lineWidth to anything, if I leave the GCodePreview.init({}) settings without the lineWidth it won't render those travel lines

Without the lineWidth parameter
ScreenShot1

With the lineWidth parameter to 1
ScreenShot2

remcoder:
Fixes #100


src/webgl-preview.ts
```diff
@@ -1,7 +1,7 @@
 import { Parser, MoveCommand, Layer } from './gcode-parser';
 import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls';
 import { LineMaterial } from 'three/examples/jsm/lines/LineMaterial';
-import { LineGeometry } from 'three/examples/jsm/lines/LineGeometry';
+import { LineSegmentsGeometry } from 'three/examples/jsm/lines/LineSegmentsGeometry';
```
sophiedeziel, line 4: In a separate branch, before actually giving a try to tube geometries, I tried to implement lines that would not depend on imports from the examples folder. My goal was to have lines that have a width relative to the model instead of relative to the screen, so zooming in would preserve the appearance of the model. I abandoned this path before I could commit a working version. 

I'm worried that this import will break in the future or make three.js upgrades risky.

remcoder: I believe these examples are quiet stable. But the dependency on them has been a nuisance. 

In the past I included the OrbitControl in the built files but it added quite a lot to the size.

remcoder: Besides, this is just a bugfix now and I was surprised it was just a wrong import. I had overlooked this earlier so I'm quite happy to discover it :-)

Issue #124, opened 2024-03-10
Proper incremental processing


remcoder:
The demo loads and renders a gcode file incrementally. But the method it uses [`processGCode`](https://github.com/remcoder/gcode-preview/blob/develop/src/webgl-preview.ts#L245) always does a complete render, for every chunk of gcode processed. This leads to a quadratic increase in total render time. This is felt with large models and with volumetric rendering. In the latter case it feels as if it grinds to a halt. 

The solution would be to only render the gcode that was just processed and keep the previously rendered parts. 

UPDATE: this is partially fixed by #136 which renders a model in an animation loop upon load.

What's still to be done is handling the case where gcode is actually _parsed_ in chunks. The problem here is that it is unknown upfront how many layers there will be eventually. This will not work well with the gradient used for line rendering, either you cannot compute the color gradient or you have to re-render the whole file. 

TODO:
- [ ] create new demo for incremental parsing & processing of gcode
- [ ] only render the new gcode 
- [ ] disable the gradient (line rendering)
- [ ] optional: find a way to re-render the model once every
- [ ] optional: pass the expected # of layers to the `processGCode` function

sophiedeziel:
I see some issues with keeping objects on the scene. As the number of meshes increases, the performance drops significantly, even for small models like the benchy. `BatchedMesh` is what optimizes the whole thing. 

The challenge with `BatchedMesh` is that we can't add geometries after the fact, as it has preallocated memory. I solved that by batching in chucks, keeping the number of objects on the scene relatively low. But we don't have the option to render line per line at the moment. If that's something we actually want, I have some ideas. 

If we're able to parse in a background worker in a way the partial results are available to the main thread, we'd be able to batch the current progress at each animation frame. The whole parsing and interpreting could be non-blocking. That way, we don't have to know anything about the size of what's parsed, we'd just render what's available as fast as the CPU is able to interpret. 

Now, using background workers in javascipt is a skill I have yet to develop and I have no idea if that's even possible.

sophiedeziel:
Did some digging and I think I have a viable approach. I'll start on a POC after we have a first alpha version released.

It would combine `requestAnimationFrame` and promises. 

Basically, all the `requestAnimationFrame` would do is the batching of available meshes and add them to the scene. As soon as the parsing starts, promises would be leveraged to distribute the heavy calculations in small units. When a promise is resolved, the result is kept available for the batching method to render on the next frame. 

[...]

PR #136, opened 2024-03-31
Fix slowdown in progressive render

remcoder:
this is a partial fix for #124 

A new method was added: `renderAnimated `. This will render the layers progressively, to show progress while loading. Meant for large files that can take many seconds to render. 

the gradient can be maintained in the DEMO because we load the whole gcode in advance. So the gcode should be parsed as a whole before rendering. For true incremental rendering the gradient should be disabled.

Render progressively: 
- [x] when initially loading benchy
- [x] when dropping a new file
- [x] when choosing a file from the file selector
- [x] render progressively when toggling tubes
- [x] maintain gradient

### DIFF
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -151,7 +151,6 @@ const settingsPresets = {
 export function initDemo() {
   // eslint-disable-line no-unused-vars, @typescript-eslint/no-unused-vars
   const settings = JSON.parse(localStorage.getItem('settings'));
-  console.debug('settings loaded', settings);
 
   const initialBackgroundColor = preferDarkMode.matches ? '#111' : '#eee';
 
@@ -406,7 +405,6 @@ export function initDemo() {
   function changeHighlightTopLayer(enabled) {
     toggleHighlight.checked = enabled;
     if (enabled) {
-      console.log('topLayerColorInput.value', topLayerColorInput.value);
       changeTopLayerColor(preview.topLayerColor || '#40BFBF');
       changeLastSegmentColor(preview.lastSegmentColor || '#ffffff');
       topLayerColorInput.removeAttribute('disabled');
@@ -559,9 +557,7 @@ async function startLoadingProgressive(gcode) {
   gcodePreview.clear();
   gcodePreview.parser.parseGCode(gcode);
   updateUI();
-  console.time('render');
   await gcodePreview.renderAnimated(Math.ceil(gcodePreview.layers.length / 60));
-  console.timeEnd('render');
   updateUI();
   startLayer.removeAttribute('disabled');
   endLayer.removeAttribute('disabled');
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -5,7 +5,7 @@ import * as Canvas2Image from 'canvas2image';
 let gcodePreview;
 let favIcon;
 let thumb;
-const chunkSize = 10000;
+let chunkSize = 10000;
 const maxToolCount = 8;
 let toolCount = 4;
 
@@ -306,7 +306,12 @@ export function initDemo() {
     fileSize.innerText = humanFileSize(file.size);
 
     preview.clear();
-
+    if (preview.renderTubes && file.size > 5 * 1024 * 1024) {
+      confirm('This file is large and may take a while to render in this mode. Change to line rendering?')
+        ? (preview.renderTubes = false)
+        : (preview.renderTubes = true);
+    }
+    preview.initScene();
     await preview._readFromStream(file.stream());
     updateUI();
     preview.render();
@@ -540,7 +545,7 @@ async function loadGCodeFromServer(file) {
 }
 
 function _handleGCode(filename, gcode) {
-  // chunkSize = gcode.length / 1000;
+  chunkSize = gcode.length / 1000;
   fileName.innerText = filename;
   fileSize.innerText = humanFileSize(gcode.length);
 
@@ -565,7 +570,9 @@ function startLoadingProgressive(gcode) {
       startLayer.removeAttribute('disabled');
       endLayer.removeAttribute('disabled');
     }
-    gcodePreview.processGCode(chunk);
+    gcodePreview.parser.parseGCode(chunk);
+    if (gcodePreview.renderTubes) gcodePreview.renderIncremental();
+    else gcodePreview.render();
     updateUI();
   }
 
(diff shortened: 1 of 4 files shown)


sophiedeziel:
I just tested it. Wow. It's so smooth now! I am impressed!

remcoder:
I finally figured out we could parse it all, know the number of layers and then render incrementally. This has the added effect that the color gradient can be accurately computed even when rendering the start. That's why it is stable now. Previously the gradient stretched out during rendering.

remcoder:
I'm not sure about the api change I did. I'm need of critical opinions

src/webgl-preview.ts
```diff
@@ -289,7 +302,7 @@ export class WebGLPreview {
     this.render();
   }
 
-  render(): void {
+  initScene() {
```
sophiedeziel: I agree with splitting out and explosing `initScene`. Not a breaking change because `render` still uses it internally. Clients don't have to explicitly call it, but they now can if they need to.

src/webgl-preview.ts
```diff
@@ -101,21 +107,27 @@ export class WebGLPreview {
...
 
+  // rendering
+  private group?: Group;
   private disposables: { dispose(): void }[] = [];
+  static readonly defaultExtrusionColor = new Color('hotpink');
   private _extrusionColor: Color | Color[] = WebGLPreview.defaultExtrusionColor;
+  private animationFrameId?: number;
+  private renderLayerIndex = 0;
+
+  // colors
```
sophiedeziel: Agree with keeping them all private (except for `defaultExtrusionColor` which can be useful to have access to, and you kept it exposed)

src/webgl-preview.ts
```diff
@@ -310,17 +323,54 @@ export class WebGLPreview {
...
       const light = new AmbientLight(0xcccccc, 0.3 * Math.PI);
       // threejs assumes meters but we use mm. So we need to scale the decay of the light
       const dLight = new PointLight(0xffffff, Math.PI, undefined, 1 / 1000);
       dLight.position.set(0, 500, 500);
       this.scene.add(light);
       this.scene.add(dLight);
     }
+  }
+
+  createGroup(name: string): Group {
```
sophiedeziel: Is this public? I think manipulating groups directly from outside the library can have unintended consequences, but can also be useful to add things to the scene. It should probably remain private until we get more confidence.

demo/js/demo.js
```diff
@@ -208,12 +216,16 @@ function initDemo() {
 
   toggleExtrusion.addEventListener('click', function () {
     changeRenderExtrusion(!!toggleExtrusion.checked);
+    preview.initScene();
     preview.render();
   });
```
sophiedeziel: `preview.render()` calls `initScene`, do you need to call it here?

src/webgl-preview.ts
```diff
@@ -310,17 +323,54 @@ export class WebGLPreview {
...
+    } else {
+      // FIXME: this is just a very crude approximation for centering
+      group.position.set(-100, 0, 100);
+    }
+    return group;
+  }
+
+  // Render the next increment. This is done by removing the previous layer and rendering all layers up to the current one.
+  // This way it preserves the state of the previous layers but it does render the top layer again b/c it probably wasnt completed before.
+  renderInc(): void {
```
sophiedeziel: If I understand correctly, the role of this method is to do a render based on state changes and the key is to use `this.prevLayerIndex` to only render what's new (and backtrack just a little in case of incomplete layers). 

There's a part I'm slightly confused. The only call is commented out, so it's technically unused.

src/webgl-preview.ts
```diff
@@ -339,6 +389,39 @@ export class WebGLPreview {
...
+
+    for (let l = 0; l < layerCount && this.renderLayerIndex + l < this.layers.length; l++) {
+      this.state = this.prevState ?? State.initial;
+      this.renderLayer(this.renderLayerIndex);
+      this.prevState = { ...this.state };
+      this.renderLayerIndex++;
+    }
+    this.scene.add(this.group);
+  }
+
```
sophiedeziel: I wonder if that should be private or not.

src/webgl-preview.ts
```diff
@@ -460,13 +543,21 @@ export class WebGLPreview {
...
+  resetState(): void {
     this.startLayer = 1;
     this.endLayer = Infinity;
     this.singleLayerMode = false;
-    this.parser = new Parser(this.minLayerThreshold);
     this.beyondFirstMove = false;
-    this.state = { x: 0, y: 0, z: 0, r: 0, e: 0, i: 0, j: 0, t: 0 };
+    this.state = State.initial;
+    this.prevState = undefined;
+    this.prevLayerIndex = undefined;
```
sophiedeziel: Same question about keeping this private. State is an internal concept that should not be messed with.

sophiedeziel (commented):
I don't want to overwhelm you with the number of comments on this PR, the overall work is really great!

I'm mostly interested in the private/public aspect, since some concepts and operations are implementation details.

src/webgl-preview.ts
sophiedeziel: I also think that the naming should tell more about that story. `renderNextIncrement` is more explicit and makes the first part of the code comment obsolete. And in the context of a library user, what is an increment? What would they understand of this API, if they have zero knowledge on how it's implemented?

remcoder: I started out the PR with this method but in the end I came up with a different solution. This is where most of my confusion some from: I still have renderInc in head as better way to do well, incremental processing. The difference is in how the parser is used. So renderInc would be used when new chunks of gcode would be parsed, processed and rendered. But for loading files in the demo, and to maintain the gradient, I needed to parse the file first.

So technically this method should be removed now. Agreed. We can always dig it up later if we need it.

src/webgl-preview.ts
remcoder: exposing it was an accident :-) I want to expose only what is required. Especially bc I think the methods initScene and clear still need to be matured

demo/js/demo.js
remcoder: good catch!

src/webgl-preview.ts
remcoder: Maybe also splitting it out to a separate class

sophiedeziel:
Are you sure we need to extract the animation in a new class right now? How would that look like?

PR #145, opened 2024-05-24
Add presets to the demo

sophiedeziel:
I know you want to rewrite the demo in Svelte, but I thought this feature would be nice.

The idea is to present the library and it's different use cases.

It also introduces logic to disable some settings that changes won't appear because of a dependency. For example, the travel color is disabled when travel is unchecked.

This is only a draft right now because I want feedback to improve the presets themselves. I think it is important to show some use cases and features, but I'm not sure all the current files represents the cases well. Maybe some diversity instead of 3 different benchies would be nice. 

Here is a starting point of the use cases to demo, for which I'd like feedback and new ideas:
- Multicolor support
- Vase mode
- CNC files
- Arc support
- Single layer mode (this one I did not implement yet because of parsing logic. It could eventually show the tool and progression like in octoprint. I digress...)
- Large files

I wonder when and why vary some of the colors, to add some variations. Also, I'm looking forward to the performance improvements so that tube rendering can be used in more cases.

I created an unnecessary large number of small functions, there must be a better way to do that.


remcoder:
great idea! Using presets for highlighting usecases would work great in the demo. 

As for the gcode files that would be appropriate, anything is fine as long as they have a permissive license and the file sizes should be kept reasonable. (AFAIK large files permanently weigh down the git repo)

sophiedeziel:
Reducing the weight of the repo would be a good reason to split out the demo out of it, do you think?

Do you have ideas on use cases and type of gcode to present? Should we also play with colors?

sophiedeziel:
I've sliced some new models!

I was mindful about the file size. I reduced the model's scale in my slicer and opted for settings that would keep the number of lines in control (thick layers, low infill, etc). 

In order to respect the model's licence, I've added attribution at the top of the resulting gcode:

- Link to the original model
- Name of the model on Thingiverse
- Name of the author
- Original licence (I chose all CC BY-NC-SA 3.0)

sophiedeziel:
Final note: I kept the default's preset exactly as they were before, but I enabled tube rendering on all other presets (that are not the CNC one). 

Looking forward to the performance improvements we're planning to have tube rendering the default!

remcoder:
Really nice work! Maybe the presets can be in a different file? demo,js is already too big. Could also be done after merging..

As for the file sizes, they're not that big actually, so for now it's fine. But yeah, we prob shouldn't have 10x the gcode files this way. Data shouldn't live in a repo anyway. We could potentially move it to cloud storage. But, not now. It's a problem when it's problem.

sophiedeziel:
Cloud storage is a very good idea! 

Agreed for the presets being in a separate file. Let's do that in a follow up, I'll be happy to take that on.

Issue #96, opened 2024-01-11
Tube rendering: dynamic radius


remcoder:
For accurate rendering using volumetric tubes (TubeGeometry) one of the first things that can be improved the radius of the tubes. For a first approximation the radius could be determined by _line width_ and _line height_.

PR #147, opened 2024-05-24
Tube's diameter set by an Extrusion Width parameter

sophiedeziel:
Part of https://github.com/remcoder/gcode-preview/issues/96

As discussed about a month or two ago on Discord, this is a first and tiny step towards dynamic tube diameter for tube rendering. 

The next steps are more complicated and require some logic and decisions that would be too much to all put in the same PR. So I'm going incrementally.  

The next step would be to try and have the layer height as well, in order to change the tube's height. With those two parameters that can be explicitly set, it unlocks fetching values from the gcode instead of hardcoded values as the default.

>[!NOTE]
> In the demo, I used the label "Tube width" instead of "Tube line width" as a shortcut to fix a display issue, it was to long for the horizontal space of the sidebar.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -65,11 +65,12 @@
           />
         </div>
         <div class="controls">
-          <label>Tube line width <output id="tube-line-width-value" ></output></label>&nbsp;<input
+          <label>Tube width <output id="tube-line-width-value" ></output></label>&nbsp;<input
             type="range"
             min="0.05"
-            max="30"
+            max="3.0"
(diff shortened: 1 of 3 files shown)


sophiedeziel:
I just re-read our thread on Discord. We aligned on `extrusionWidth` for the naming. 

I'll make that change later today.

remcoder:
Awesome you're working on this!

remcoder:
🎉

PR #149, opened 2024-05-26
ExtrusionGeometry: use a custom geometry for extrusions

sophiedeziel:
Addresses https://github.com/remcoder/gcode-preview/issues/95
Also touches https://github.com/remcoder/gcode-preview/issues/96 by introducing line height. 

What we've planned for so long is finally here!

A custom geometry that fits better our needs!

The main difference between the standard `TubeGeometry` and this `ExtrusionGeometry` is how segments are distributed along the path.

Instead of using equidistant points on a curve, the `ExtrusionGeometry` places the frames at each change of direction of the extrusion moves. Long straight lines are no longer "hoarding" segments and corners are sharper. Performance is also improved because fewer segments are needed for the same paths.

A custom geometry also allows to implement line height and line width the way we want. So I went ahead and added it right away.

There's still work to be done before it is ready to merge. 

- [x] Cleanup old prototyping comments
- [x] Convert the geometry file to typescript
- [x] Refactor `generateSegment` to make it easier to read
- [x] Add some tests

Here's some comparison between the `develop` branch and this branch:

| `develop` | `feature/custom-volume` |
|--------|--------|
| Screenshot 2024-05-30 at 00 32 54 | Screenshot 2024-05-30 at 00 31 50 |
| Screenshot 2024-05-30 at 00 41 04  | Screenshot 2024-05-30 at 00 39 56 |


remcoder:
Again, really nice work! This will be a huuuge improvement and I am sure we can start using the volumetric rendering per default when this is merged. 

So if I'm correct, for the benchy there seems to be a reduction in triangles of about 45% yielding a speedup of about 2x. That's awesome! 

I'll do some testing too to see how this translates to different models.

remcoder:
I tested with a 13MB file and it seems to only take 2x the time to render compared to simple lines! That's freakin' amazing!

sophiedeziel:
I am myself impressed by what you just tested. I did not expect that large of an improvement!

And did you notice the improvement of the render quality too? 😸

remcoder:
Honestly no. But I did notice the number of radial segments is now 8 instead of 4. Is that what you mean?

sophiedeziel:
The default is the same as the initial TubeGeometry, but WebGLPreview specifies 4.

The improvement in quality is not related to the number of radial segments, it has an overall very small impact on a whole part. What's improved is the placement of segments and the removal of the curve interpolation. Corners are sharp because they are accurate!

sophiedeziel:
Just took some quick screenshots to show you:

# Before
image

# After
image

remcoder:
Oooh, nice! How did miss that! Nice improvement :-)

Issue #153, opened 2024-05-30
Dev mode


sophiedeziel:
I'm thinking about adding a dev mode to display some informations on the screen that would help the development of features. 

Just a rectangle on the top right corner to show some stats and toggle some dev features.

- Time to render
- Size of the gcode
- Number of travel moves
- Number of extrusion moves
- Number of triangles on the scene
- Number of groups
- Camera position
- Toggle wireframe mode
- Toggle some `console.log` outputs

This can be done incrementally

remcoder:
Great idea!

Would be useful for non-dev users too.

PR #157, opened 2024-06-01
Introduce Dev mode

sophiedeziel:
# Context
Part of #153 

I started some explorations on how I could implement the dev mode. 

Screenshot 2024-06-01 at 12 42 24

## Approach
The easiest path I found was using the lil-gui library included in Three.js.

There are many advantages to that library:

- No need to add dependencies
- Neat API that allowed me to add controls in very few lines of code
- Built to control attributes, so it's easy to play with them
- No need to style ourselves (but we can if we want)
- Built-in hiearchy that helps organize the information
- Many control types built-in

# Discussion

## Do you think the demo could use the library as well? 
That will simplify the demo code a lot, IMO. We can style the panel so it's prettier for demo purposes.

### DIFF
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -160,7 +160,8 @@ export function initDemo() {
     buildVolume: settings?.buildVolume || { x: 190, y: 210, z: 0 },
     initialCameraPosition: [180, 150, 300],
     backgroundColor: initialBackgroundColor,
-    lineHeight: 0.3
+    lineHeight: 0.3,
+    devMode: { camera: false, renderer: true }
   }));
 
   backgroundColor.value = initialBackgroundColor;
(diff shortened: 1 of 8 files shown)


remcoder:
Seems like a great choice to use lil-gui. And yes It could be part of the lib, especially since it is based on threejs so it won't add new deps.

sophiedeziel:
That's what I was thinking. Having debug and dev tools helps in the library's development, but also helps development of apps using and interacting with the library.

remcoder:
Totally agree!

sophiedeziel:
That's weird, the preview link does not show the demo mode. I'm not sure if I messed up the branch update or if it's deploying the develop branch. Looking at the source, it looks like develop.

sophiedeziel:
Fixed the preview link in https://github.com/remcoder/gcode-preview/pull/162

remcoder:
Just played with the dev tools. This is sooo useful! Being able to see the # of triangles, or change the camera position. Love it!

Maybe the buttons to store/restore the camera position we talked about should go in there? If possible?

sophiedeziel:
> Maybe the buttons to store/restore the camera position we talked about should go in there? If possible? 

Yes! Entirely possible!!! And so easy!

PR #158, opened 2024-06-02
Module imports for demo

### LINKED ISSUE #97: Update demo to use the jsm module system
THREE.js no longer provides an iife build ( >= r160). This means we'll have to support the new module build and update the demo accordingly. 

[more info](https://threejs.org/docs/#manual/en/introduction/Installation)

sophiedeziel:
Closes https://github.com/remcoder/gcode-preview/issues/97

I tried updating to three.js v165 but the render of the lines did not work properly. So that will need to be fixed separately.


remcoder:
awesome!

could the import map be changed into a single line with a `script` tag and `type=module`? Would that still work?

sophiedeziel:
>could the import map be changed into a single line with a script tag and type=module? Would that still work?

I can try to make it shorter, maybe by splitting in it's own file. But my understanding is that the `importmap` type is a specific type and I'm not sure how to make this work with a `type=module` instead.

sophiedeziel:
I tried 
```html
<script type="importmap" src="importmaps.json"> </script>
```

and got

Screenshot 2024-06-03 at 17 21 54

I dug deeper and found: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap

Screenshot 2024-06-03 at 17 27 11

PR #182, opened 2024-06-29
Fix/stats counter position

### LINKED ISSUE #176: FPS counter position
The FPS counter is positioned awkwardly overlapping the name of the app.

image

remcoder:
changes 

- Adds `opts.devMode.statsContainer`, for passing an html element that will be the container for the stats object.
- Adds the classname `stats` to facilitate styling.
- Only instantiate the stats object when devMode is active

fixes #176

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -27,10 +27,10 @@
 
   <div class="dnd">☝️ Drop a gcode file here to view it</div>
   <div class="version-box">
-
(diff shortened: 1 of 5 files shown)


remcoder:
@sophiedeziel since you've created the devMode, you might have an opinion on my changes. I was wondering if the new option (`statsContainer`) should be in the devMode object. And if not, wether the devMode and stats should really be separate..

sophiedeziel:
No strong opinion, but I lean towards an option as part of the dev mode options. 

It could be part of the devgui panels, I think that would be the best for a unified dev experience

PR #184, opened 2024-07-27
Poc: demo in sveltekit / tailwind / FlowBite

remcoder:
This is a poc to explore options for rebuilding the demo UI using modern frameworks. It uses SvelteKit and TailwindCSS + FlowBite. While I like 'm I'm not sure they're ideal for this usecase. I would like to explore other options so this not meant to be merged.

NOTE: screenshot is for reference. The poc is more about the technical setup

image

note: the generated preview below doesn't show the changes. You can find the poc by doing a local checkout and see the svelte-demo dir


remcoder:
I'm not going through with this.

PR #185, opened 2024-07-27
Poc: demo in nextjs / bulma

remcoder:
this is another poc for rebuilding the demo UI. 

It uses Next.js, and therefore React. It uses Bulma as a base for styling.

NOTE: the screenshot is for reference. The poc is more concerned with the technical setup. 

TODO
 - [ ] render gcode preview

image


remcoder:
I'm not going through with this.

PR #169, opened 2024-06-23
WIP Refactor responsibilities out of WebGLPreview

sophiedeziel:
This is a POC for splitting the codebase into smaller classes

PR #188, opened 2024-08-13
Extract out BuildVolume

sophiedeziel:
As explored in https://github.com/remcoder/gcode-preview/pull/169, I've extracted out the build volume to encapsulate the attributes and geometry building.


remcoder:
I think it's fine for now, separating it out as a util. Eventually we'd want to move to a scheme where we add/remove objects like a build plate to the scene. 

Also it might be better to separate the axes from the build plate, eventually that is.

sophiedeziel:
Agreed, they could all be separated. In a way, I find it elegant to have the build volume in a single object but with different methods to represent it on the scene. They all use the same x,y and z values.

remcoder:
@sophiedeziel this LGTM. Agree to merge?

sophiedeziel:
Yes!

PR #196, opened 2024-09-16
Handle start/end layer changes gracefully

remcoder:


remcoder:
Closing this as the side-effect isn't as straightforward as expected. For instance, when setting both props immediately after each other, the effect of the fist prop on the second is negated. This is problematic in scenarios where all settings are set at the same time like when using a view model object to hold settings for the UI.

PR #183, opened 2024-07-16
feature: different travel color to distinguish air-cutting and actual-cutting in CNC

xuzhen1994:
different travel color to distinguish air-cutting and actual-cutting in CNC.

PR #211, opened 2024-10-09
Introduce an interpreter to extract command interpretation logic

sophiedeziel:
This refactor is extracting responsibilities out of the parser and out of the webgl-preview class into an interpreter to produce job and paths primitives. The line used to be blurry in between the two classes. The interpreter is the missing piece that helps defining the boundaries in a clear way. 

# Key concepts

## The parser
The parser's responsibility is only about converting a text file to an array of structured javascript objects that represents the list commands and their attributes. All the parser cares about is the syntax. It should never hold logic related to what the commands means and how they should be interpreted. If state has to be held based on the content of the gcode, it's a smell about code not belonging in there.

## The interpreter
The interpreter takes the output of the parser. This is where the gcode commands are "implemented". It manages a virtual machine's state and generates a list of paths representing extrusions and travels. All it should care about is translating gcode into job and path primitives.

## The Path
Represents a continuous travel of the tool, with specific attributes and type. It holds the coordinates and the logic to be rendered on a scene.

## The job
Represents the state of a print or milling job. It holds info about current state and all the paths generated by the interpreter "running" the gcode on it. 

## The renderer
The webgl-preview class is the renderer. Maybe it should be renamed `renderer`? It takes the state of the virtual machine and the results from the interpreter and adds them to a 3D scene.

# Next steps

This PR is still a work in progress and there are a lot of things I am not satisfied with yet. 

For example, I've put 3D objects logic into the paths objects. It does not seem right to me. It makes the materials management overly complicated. 
I'm also annoyed by the gcode enum for supported commands. It should be simpler to implement a new command. 
For some reason, it is not very performant at the moment. This will be investigated after the refactor gets into a more stable state.

## Todo
- [x] Reimplement the progressive rendering
- [x] Make sure GCode can be processed in chunks
- [x] Fix existing tests
- [x] ~Debug the CNC toolpath preset that does not render correctly~ turns out it's [an issue on `develop`](https://github.com/remcoder/gcode-preview/issues/216), it was not introduced by the changes of this PR. 
- [x] Extract out new features that creeped in (`G21`)
- [x] Extract out a PR for the demo fix for progressive rendering

## Nice to have
- [x] Add tests for the `Interpreter` class
- [x] Add tests for the `Path` class
- [x] Add tests for the `Job` class

## Not in scope for this PR
This PR is already large enough. Some shortcuts and less than ideal implementations were used to minimize the size of it. Listing them here as an acknowledgment of them and we'll open PRs later to address them individually.

- Refactor to parse and interpret outside of the renderer
- Improve the webgl-preview tests
- Improve the methods filtering/looping through all commands or paths (https://github.com/remcoder/gcode-preview/pull/220)
- Improve the GCode Enum situation
- Simplify parsing the GCode commands (`g1`/`g01` and all the `T`)
- Bring back layering split with tolerance


src/gcode-parser.ts
```diff
@@ -128,22 +174,23 @@ export class Parser {
...
   } {
     const lines = Array.isArray(input) ? input : input.split('\n');
 
     this.lines = this.lines.concat(lines);
 
-    const commands = this.lines2commands(lines);
+    this.commands = this.lines2commands(lines);
 
-    this.groupIntoLayers(commands);
+    // this.groupIntoLayers(this.commands);
```
sophiedeziel: I am thinking about letting the interpreter group into layers, but to not have rendering logic looping on layers by default. This would allow more flexibility for rendering non-planar gcodes while still be able to support single layer mode and min/max layers.

src/gcode-parser.ts
```diff
@@ -60,13 +60,69 @@ type MoveCommandParamName = 'x' | 'y' | 'z' | 'r' | 'e' | 'f' | 'i' | 'j';
...
+      case 't5':
+        return Code.T5;
+      case 't6':
+        return Code.T6;
+      case 't7':
+        return Code.T7;
+      default:
+        return undefined;
+    }
+  }
```
sophiedeziel: There must be a better way. The idea here is to parse the text gcode command into a predictable format. I used an enum to make it easy to type, but there's something smelly about this. Any idea on how to change that?

src/interpreter.ts
```diff
@@ -0,0 +1,176 @@
...
+    commands.forEach((command) => {
+      if (command.code !== undefined) {
+        this[command.code](command, machine);
+      }
+    });
+
+    return machine;
+  }
+
+  G0(command: GCodeCommand, machine: Machine): void {
```
sophiedeziel: Each method must be named after the enum for the command and accept those two params. There's probably a better way to do this.

src/interpreter.ts
```diff
@@ -0,0 +1,176 @@
+import { Path, PathType } from './path';
+import { Code, GCodeCommand } from './gcode-parser';
+import { Machine } from './machine';
+
+export class Interpreter {
```
sophiedeziel, line 5: It would be cool for lib users to be able to use their own custom interpreter (or override this one). The [features around distances on the CNC features PR](https://github.com/remcoder/gcode-preview/pull/183/files#diff-8ec787c979ef6cb88e426e798a75cc41d326888eaca61c5f5691b18c5e65c729R174-R181) could be possible as a custom interpreter instead of an official part of this library. 

There could also be some slicer specific interpreter dictionaries implemented in the interpreter to be able to implement some preview features more easily. For example, comments to mark extrusion types (infill, outer wall, wipe) are different depending on the different slicers.

src/interpreter.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    state.x = x || state.x;
+    state.y = y || state.y;
+    state.z = z || state.z;
+
+    lastPath.addPoint(state.x, state.y, state.z);
+  }
+
+  G1 = this.G0;
```
sophiedeziel, line 34: Aliases!

src/machine.ts
```diff
@@ -0,0 +1,66 @@
...
+  units: 'mm' | 'in';
+
+  static get initial(): State {
+    const state = new State();
+    Object.assign(state, { x: 0, y: 0, z: 0, e: 0, tool: 0, units: 'mm' });
+    return state;
+  }
+}
+
+export class Machine {
```
sophiedeziel: It used to be called `Print`, but the name does not apply to the CNC case. 

I'm not sure about the `Machine` name but it's the best I could come up with so far. It is supposed to represent the state and result after running some GCode.

src/machine.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  isPlanar(): boolean {
+    return (
+      this.paths.find(
+        (path) =>
+          path.travelType === PathType.Extrusion && path.vertices.some((_, i, arr) => i % 3 === 2 && arr[i] !== arr[2])
+      ) === undefined
+    );
+  }
```
sophiedeziel: This may be an implementation that's too simple. If there is at least one extrusion line that has vertical movement, the print is not considered planar.

src/machine.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  layers(): Path[][] | null {
+    if (!this.isPlanar()) {
+      return null;
+    }
```
sophiedeziel: I kinda like grouping layers together only when they are actual layers. 

Right now, it forces me to decouple the rendering from the layers. It breaks some of the nice features we have, but they most likely can be implemented differently.

src/path.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      this.geometryCache = new ExtrusionGeometry(this.path(), this.extrusionWidth, this.lineHeight, 4);
+    }
+    return this.geometryCache;
+  }
+
+  line(): BufferGeometry {
+    return new BufferGeometry().setFromPoints(this.path());
+  }
```
sophiedeziel: `geometry()` and `line()` are the implementation of the two render modes. I don't think it should be implemented here. I think it should be a specific render class sharing a common abstract interface that `webgl-preview` can interact with, passing paths and receiving something that can be added to a scene. The cache implemented here is useful to not recalculate the whole geometry each time, but that's a premature optimization here. I'm also concerned about material management

src/interpreter.ts
sophiedeziel: > For example, comments to mark extrusion types (infill, outer wall, wipe) are different depending on the different slicers.

Actually, that should be in the parser and formatted in a standard way for the interpreter to interpret.

src/gcode-parser.ts
remcoder: hmmm, instead of Enums, union types of string literals could be used too: 
`type Command = 'g0' | g2 | .. `

It's a little cleaner but suffers from the same problems. 

Another way is to us `as const`:
`const commands = ['g0', 'g1', ...] as const` 

This creates a type from the whole array literal. But since it already is a value we can use it to compare other values against, and use functions like toLowercase to allow for variants. I think it is the cleanest and DRYest way in terms of bothe types and run-time code.

remcoder: We could also interpret `t` as the command and the number after it as a param. I was just lazy I guess when I first wrote it ;-)

src/gcode-parser.ts
remcoder: For min/max layer I plan to use clipping planes. This means we can change the min/max without having to rerender the 3d object.

So layer grouping can be removed from the parser. 
For now, the layer height is a constructor param but eventually we want the parser to read this from the gcode.

src/interpreter.ts
remcoder: I can't think of any :-(

src/interpreter.ts
remcoder: Do you think we'll get away with a customizable parser but a generic interpreter? I think for the most part yes but I trying to come up with a counter example... can't think of one atm..

The parser could be a constructor param (injection) and if not present, we instantiate the default parser. This would be backwards compatible, so it could be a v2 feature.

src/interpreter.ts
remcoder: aha! nice!
this can even be moved up, before the method declaration, so we can have all aliases together and visible

src/machine.ts
remcoder: Also looks kinda expensive. Should we do this check in the parser?

src/machine.ts
remcoder: Analogous to `Print` but more generic would be `Job`. That's also what Octoprint calls it. How does that sound?

src/machine.ts
remcoder: for v3 we could start w/o layers but if you want to merge the interpreter into v2 we still need the layers

src/machine.ts
```diff
@@ -0,0 +1,66 @@
...
+          currentLayer = [];
+        }
+        currentLayer.push(path);
+      }
+    });
+
+    return layers;
+  }
+
+  extrusions(): Path[] {
```
remcoder: this looks very expensive. It will filter all paths every time and allocate a subset.

src/path.ts
remcoder: Great idea to have a Path class!

src/interpreter.ts
sophiedeziel: I think that anything related to calculating (the distance for example) should be in the interpreter. 

And yes, I'm thinking about injection/default, for both the parser and the interpreter. There's value into having custom in all cases, as long as the AST is solid

src/machine.ts
sophiedeziel: `Job` is fantastic. Yes!

src/machine.ts
sophiedeziel: It is, but it's only ran once in the flow. This is a first lazy implementation before I consider adding an index approach that would work better for different rendering modes.

src/machine.ts
sophiedeziel: If all the parser cares about is the syntax of gcode, this logic does not belong in the parser. However, if the gcode has a special comment about layers, that's another story. 

The role of the interpreter is to encapsulate anything that has to do with understanding what is actually going on. 

I had another implementation that had some tradeoff I did not want to make. It was directly in the G1 method. It had the advantage of not looping a 2nd time through all commands. But also had the disadvantage to add a condition no matter if it was already determined to be non-planar. 

I did not implement memoization because of processing gcode in chunks, but could be a short term solution.

src/gcode-parser.ts
sophiedeziel: You just gave me an idea. Any type of command is a letter + a number. This could help with parsing `g1`/`G01`, if we decompose it.

src/gcode-parser.ts
sophiedeziel: I don't think we should remove layers completely. We'll have interesting challenges when comes the time to support adaptive layers properly! But right now, we're able to completely decouple rendering from layers. I'm looking forward to see the implementation using clipping. It will be so smooth.

src/path.ts
sophiedeziel: Yes, it unlocks so many things!!

src/machine.ts
sophiedeziel: I'm aiming for v3, but breaking changes still need a good justification. Layers are very useful and lib users have a good chance of using them.

src/interpreter.ts
sophiedeziel: Maybe it's a good thing to not be flexible?

remcoder:
Nice work @sophiedeziel! So many improvements!

Some dev process ideas that I want to throw out there:
- most importantly I'd like to keep the diff for every PR not larger than needed. Focused PRs. This allows me for instance to pick up the PR later. 
- the main focus was to introduce  the Interpreter. This could be not much more than moving code around since the logic is already present.
- this means that other improvements really deserve there own PR. 
- smaller PRs are so nice because they create a more consistent flow of smaller improvements and prevent getting stuck on large changes. 

At this point it would seem we can create a a separate PR for parser improvements. And another for splitting of a Renderer class. 

And then there can be PRs for improving logic or perf or cleanup. 

Most of this, except for the removal of layers, can all still be merged into V2. 

I know we said we were ready for V3 but I think now that we would be in a better place  if we tackle some things like this refactor first.

Your thoughts?

sophiedeziel:
It is such a challenge to keep small! I agree with you that some performance stuff should be left for later. But there are some breaking changes that are not simply about moving code around, unless we introduce some of the classes in an unused state in parallel with the old code.

This PR should at a minimum be in a working state, which it is not 100%. 

It should also include some minimal tests to help us in the next iterations. 

Are you able to identify a part that could be it's own individual PR? That could be a start. 

I'll scope out some of the todos, to prioritize mergability.

And how strongly do you feel about progressive rendering?

src/machine.ts
sophiedeziel: As discussed in discord, optimisations will come in a follow-up PR

src/machine.ts
sophiedeziel: As discussed in Discord, this PR is directly for v3.x and will merge in the `v3.x` branch. 

However, there should at least have an implementation using the tolerance logic that's in the current version. Will follow-up with a PR addressing that.

src/__tests__/interpreter.ts
```diff
@@ -0,0 +1,269 @@
...
+  const command = new GCodeCommand('G0 X1 Y2', 'g0', { x: 1, y: 2 });
+  const interpreter = new Interpreter();
+
+  const job = interpreter.execute([command]);
+
+  expect(job.paths.length).toEqual(1);
+  expect(job.paths[0].vertices.length).toEqual(6);
+  expect(job.paths[0].vertices[0]).toEqual(0);
+  expect(job.paths[0].vertices[1]).toEqual(0);
+  expect(job.paths[0].vertices[2]).toEqual(0);
```
remcoder, line 81: Maybe add test for vertices 3..5 here?

src/__tests__/interpreter.ts
```diff
@@ -0,0 +1,269 @@
...
+  const command = new GCodeCommand('G0 X1 Y2', 'g0', { x: 1, y: 2 });
+  const interpreter = new Interpreter();
+  const job = new Job();
+
+  interpreter.G0(command, job);
+
+  expect(job.paths.length).toEqual(1);
+  expect(job.paths[0].travelType).toEqual('Travel');
+});
+
```
remcoder, line 133: shall we add a test for negative extrusion (retraction)? It should be interpreted as travel.

src/__tests__/interpreter.ts
```diff
@@ -0,0 +1,269 @@
...
+  const job = new Job();
+  interpreter.execute([command1], job);
+
+  interpreter.G0(command2, job);
+
+  expect(job.paths.length).toEqual(1);
+  expect(job.paths[0].vertices.length).toEqual(9);
+  expect(job.paths[0].vertices[6]).toEqual(3);
+  expect(job.paths[0].vertices[7]).toEqual(4);
+  expect(job.paths[0].vertices[8]).toEqual(0);
```
remcoder: This implies the state was initialised at z = 0. But isn't this test really about tate transition? So I rather suggest an explicit initialisation: `job.state.z = 42` to demonstrate that z is left unchanged because this cmd has no z param.

src/__tests__/interpreter.ts
```diff
@@ -0,0 +1,269 @@
...
+
+test(".G0 assigns the travel type if there's no extrusion", () => {
+  const command = new GCodeCommand('G0 X1 Y2', 'g0', { x: 1, y: 2 });
+  const interpreter = new Interpreter();
+  const job = new Job();
+
+  interpreter.G0(command, job);
+
+  expect(job.paths.length).toEqual(1);
+  expect(job.paths[0].travelType).toEqual('Travel');
```
remcoder: suggestion to reference the enum in the test: `.toEqual(TravelType.Travel)`.

src/__tests__/interpreter.ts
```diff
@@ -0,0 +1,269 @@
...
+  const command2 = new GCodeCommand('G1 X3 Y4 E5', 'g1', { x: 3, y: 4, e: 5 });
+  const interpreter = new Interpreter();
+  const job = new Job();
+  interpreter.execute([command1], job);
+
+  interpreter.G0(command2, job);
+
+  expect(job.paths.length).toEqual(2);
+  expect(job.paths[0].travelType).toEqual('Travel');
+  expect(job.paths[1].travelType).toEqual('Extrusion');
```
remcoder, line 167: Paths are so cool!!! :D

src/__tests__/job.ts
```diff
@@ -0,0 +1,164 @@
+import { test, expect, describe } from 'vitest';
+import { Job } from '../job';
+import { PathType, Path } from '../path';
+
+test('it has an initial state', () => {
+  const job = new Job();
+
+  expect(job.state.x).toEqual(0);
```
remcoder: If you check `expect(job.state).toEqual(State.initial)` you ensure an exact match with neither having a prop the other doesn't have.

src/__tests__/interpreter.ts
```diff
@@ -0,0 +1,269 @@
...
+});
+
+test('.execute runs on an existing job', () => {
+  const job = new Job();
+  const command = new GCodeCommand('G0 X1 Y2 Z3', 'g0', { x: 1, y: 2, z: 3 });
+  const interpreter = new Interpreter();
+
+  const result = interpreter.execute([command], job);
+
+  expect(result).toEqual(job);
```
remcoder, line 54: I think this check need to use `toBe` which checks for identity, not just equality.

src/path.ts
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable no-unused-vars */
+import { BufferGeometry, Vector3 } from 'three';
+import { ExtrusionGeometry } from './extrusion-geometry';
+
+export enum PathType {
+  Travel = 'Travel',
```
remcoder, line 6: Do we need to specify the value? If we don't it's just 0 and 1. And that't fine. We should only reference the values as the enum anyway, never as the actual value.

src/path.ts
```diff
@@ -0,0 +1,61 @@
...
+  Extrusion = 'Extrusion'
+}
+
+export class Path {
```
remcoder, line 10: great thinking ahead with the inclusion of `extrusionWidth` and `lineHeight` props! variable width and height is within reach 💪🏻

src/interpreter.ts
```diff
@@ -0,0 +1,168 @@
...
+    const { state } = job;
+
+    let lastPath = job.paths[job.paths.length - 1];
+    const pathType = e ? PathType.Extrusion : PathType.Travel;
+
+    if (lastPath === undefined || lastPath.travelType !== pathType) {
+      lastPath = this.breakPath(job, pathType);
+    }
+
+    state.x = x || state.x;
```
remcoder: this goes wrong when x,y or z are 0, which is falsy. Use ?? to handle undefined: `state.x = x ?? state.x`.

src/__tests__/path.ts
```diff
@@ -0,0 +1,129 @@
...
+  const path = new Path(PathType.Travel, 9, undefined, undefined);
+
+  path.vertices = [0, 0, 0, 1, 2, 3];
+
+  const result = path.geometry() as ExtrusionGeometry;
+
+  expect(result.parameters.lineWidth).toEqual(9);
+});
+
+test('.geometry returns an ExtrusionGeometry with the path line height', () => {
```
remcoder, line 90: Honest question: what are you testing here? I would expect that this is covered by the type system. As long as we're using the type checker I don't think you end up breaking this. Or am I missing something?

src/webgl-preview.ts
```diff
@@ -403,13 +304,20 @@ export class WebGLPreview {
     this.initScene();
 
     this.renderLayerIndex = 0;
+
+    if (this.job.layers() === null) {
```
remcoder, line 308: I believe you're executing `job.layers()` multiple times within a single render pass. You prob want to avoid that by just keep a local variable or instance prop _layers or something.

src/webgl-preview.ts
```diff
@@ -423,130 +331,21 @@ export class WebGLPreview {
...
-        this.state.z = next.z;
-        // if (next.e) state.e = next.e; // where not really tracking e as distance (yet) but we only check if some commands are extruding (positive e)
-        if (!this.beyondFirstMove) this.beyondFirstMove = true;
-      }
-    }
-
-    this.doRenderExtrusion(currentLayer, index);
-  }
+    const endIndex = Math.min(this.renderLayerIndex + layerCount, this.job.layers().length - 1);
+    const layersToRender = this.job
```
remcoder: isn't this a list of paths, now that you've flattened the layer? suggestion to name it pathsToRender

src/webgl-preview.ts
```diff
@@ -772,47 +398,64 @@ export class WebGLPreview {
...
+          color = this._extrusionColor.getHex();
         }
-      }
+
+        this._geometries[color] ||= [];
+        this._geometries[color].push(path.geometry());
+      });
     }
+
+    for (const color in this._geometries) {
```
remcoder, line 450: brilliant!

remcoder (approved):
This setup is a major improvement in many aspects. Indeed this is becoming a 'platform' we can build the next iteration of cool features on! I'm genuinely grateful for having such a brilliant contributor!! thank you for your hard work! 

I did find some things that might deserve attention but nothing that stands in the way of merging this PR. I suggest your read through my remarks and then merge at will.

src/__tests__/interpreter.ts
sophiedeziel: Good point! Is that the behavior of the current version that I missed or something to add?

src/path.ts
sophiedeziel: I find it easier using the debugger to have values. 0s and 1s are easier on the memory, but right now it's soooo useful in the development flow. 

I suggest changing that when Paths are more mature and the code is more stable.

src/path.ts
sophiedeziel: I just realized that the implementation is not complete. Chanhing the extrusingWidth in the demo has no effect! 

Follow-up PR or an additional commit here?

src/__tests__/path.ts
sophiedeziel: I was wrestling with this test and other ones that are similar. 

I wanted to make sure the geometry was passed the right attributes to the constructor. Sometimes it's valuable, sometimes it's not.

src/webgl-preview.ts
sophiedeziel: This is fixed in https://github.com/remcoder/gcode-preview/pull/220 🎉

src/__tests__/interpreter.ts
remcoder: that was existing :D
https://github.com/remcoder/gcode-preview/blob/develop/src/webgl-preview.ts#L473

src/webgl-preview.ts
remcoder: ok nvrnd, I've seen the indexers PR now :-)

src/webgl-preview.ts
sophiedeziel: This is very similar to what we have in v2. Current users already benefit from it!

src/__tests__/interpreter.ts
sophiedeziel: Perfect! I'll make sure it's still the case

src/interpreter.ts
sophiedeziel: Thanks for pointing it out! In Ruby, `0 || 8` returns 0!

PR #222, opened 2024-10-13
Don't recalculate Geometries at each render

sophiedeziel:
With every re-render, geometries are recalculated even though paths have not changed. Memoizing the geometry building method helps with that. 

It has no perceptible performance impact the initial render, even with progressive rendering. However, the next renders are significantly faster and we can notice.

There are downsides of memoizing as it requires more memory. It should be an option exposed, because it really depends on the use case. If the same model has to be re-rendered multiple times, memoization is great. If only one render is needed per gcode file, cache should be off. I am not sure what the default should be. 

What do you think? Should this PR include the option? What should be the default?

### DIFF
--- a/src/path.ts
+++ b/src/path.ts
@@ -13,6 +13,8 @@ export class Path {
   extrusionWidth: number;
   lineHeight: number;
   tool: number;
+  _geometry: BufferGeometry | undefined;
+  _line: BufferGeometry | undefined;
 
   constructor(travelType: PathType, extrusionWidth = 0.6, lineHeight = 0.2, tool = 0) {
     this.travelType = travelType;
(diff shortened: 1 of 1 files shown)


sophiedeziel:
I took time to think about it. This is a net new feature that does not come without risk. 

This idea seemed easy an quick, but I'm realizing it will be a distraction and a slow down to get v3.x in a mergable state. Changing any setting that affects the geometry (`extrusionWidth` and `lineHeight`) requires cache invalidation and we should not embark in that right now. 

I suggest to bring that optimization back later on, as an experimental feature that is off by default. The risk of introducing bugs or having unintended effects is pretty high. 

Closing this PR with the intention to revisit the idea later if we want to.

PR #237, opened 2024-10-18
Fix thick lines and set min width to 1

### LINKED ISSUE #216: Some lines are not rendered properly
The only way I found to reliably reproduce is using the CNC preset, but it also happens sometimes when rendering travel lines.

Screenshot 2024-10-11 at 19 30 24

It is hard to see, but it seems like the lines are on the scene. They are just so small we can't see them. We can notice some green pixels.

remcoder:
fixes #216 

Apparently the lineWidth for 'fat lines' in threejs was changes in r165.


sophiedeziel:
I love that this fix comes with the cleanup of one of the two different ways of rendering lines!

PR #250, opened 2024-11-11
STL export

### LINKED ISSUE #249: Add STL export button
Based on an idea by @merlin-dev-33:

What: The idea is to be able to export the currently viewed model as a 3d object. So from gcode to STL for instance. 

Why: The reason is to use the tube rendering to create a new version of the model. One application would be to first slice a model with exaggerated layer line height/width, import it into gcode-preview, export the generated mesh consisting of tubes, and the slice and print on a regelar printer with normal sized nozzle. 

How: There is an STL Exporter included in Threejs. It requires access to the Scene or the Mesh that is to be exported. The Scene is already exposed by the lib but the 3DObject that represents the loaded model isn't exposed explicitly so we need to traverse the Scene or expose the model explicitly. 

V3: Since we have an alpha for the new v3 version ready let's build this on top of that. No new dev work should be done on v2.

remcoder:
this will implement #249 

it's a work in progress. Still throws an error atm


remcoder:
I'm not sure there is much gain in getting this merged. Leaning towards closing..

remcoder:
Closing as dormant — no activity since 2024 and current effort is focused elsewhere. Can be re-opened later if there's renewed interest.

PR #270, opened 2024-12-30
Add tsdocs for everything and improve typings for readability

remcoder:
This PR adds a lot of docs to the parser. And it reworks some types for readability. Workings remain the exactly the same.

Added tsdocs for everything. Typedoc now generates more useful documentation.


src/build-volume.ts
```diff
@@ -3,20 +3,39 @@ import { AxesHelper, Group, Vector3 } from 'three';
...
+  /** List of disposable objects that need cleanup */
   private disposables: Disposable[] = [];
 
+  /**
+   * Creates a new BuildVolume instance
+   * @param x - Width in mm
+   * @param y - Depth in mm
+   * @param z - Height in mm
+   * @param color - Color for visualization (default: 0x888888)
+   */
```
sophiedeziel, line 27: There is also an inch mode (that does not impact anything really, except some arc support calculations). It is true that it's `mm` by default, but not necessarily.

sophiedeziel (commented):
I know this is already merged, I was really curious about what docs were generated on your first pass!

Something small drew my attention, here's some feedback for a next iteration.

PR #275, opened 2025-01-13
Update color instantly

remcoder:
This is an improvement to change the color of the model instantly, without having to rebuild the mesh.

It uses a simple ShaderMaterial which has an updatable uniform property for the color. Updating this doesn't require rebuilding the mesh, it only requires a renderer.render call.

It actually works better than expected so I promoted this from poc to pr.

TODO
- [x] multicolor
- [x] bring back clipping planes
- [x] ~fix drag n drop~ separate issue #277 

I needed a good model to test shader with so I included Marlin (by cipis)
image
It looks great with this shader!


src/dev-gui.ts
```diff
@@ -113,13 +112,17 @@ class DevGUI {
...
       this.saveOpenFolders();
     });
-    render.add(this.watchedObject.renderer.info.render, 'triangles').listen();
-    render.add(this.watchedObject.renderer.info.render, 'calls').listen();
-    render.add(this.watchedObject.renderer.info.render, 'lines').listen();
-    render.add(this.watchedObject.renderer.info.render, 'points').listen();
-    render.add(this.watchedObject.renderer.info.memory, 'geometries').listen();
-    render.add(this.watchedObject.renderer.info.memory, 'textures').listen();
-    render.add(this.watchedObject, '_lastRenderTime').listen();
+    render.add(this.webglPreview.renderer.info.render, 'triangles').listen();
```
sophiedeziel, line 115: Be careful with changing `watchedObject`. I had issues with this referencing some old objects when they are replaced. 

We have many `this.an_object = new ObjectClass()` that made it a challenge to watch on the GUI.

remcoder: Thanks for the heads-up. I doublechecked and I don't think this changes breaks anything. The watchedObject is still the same instance but now it is typed to make the compiler happy.

sophiedeziel: Have you checked with the other places where `watchedObject` is changed too?

remcoder: I'll check it again! Will try tomorrow.

remcoder: I checked all the dev helpers. Nothing is broken by this PR. 

The Job info and _lastRenderTime don't update but I checked that was already the case be for this PR.

Only the types have changed and the namei of the instance but there shouldn't be even a change in the javascript code that is being generated. 

btw, did you see you can now change the lighting through the dev helper?

PR #286, opened 2025-01-29
Set a default value for lineHeight

### LINKED ISSUE #285: Line rendering fails with an error
Hello,

Since last night, I have been receiving an error message when rendering GCode files, like the one I have attached to this ticket (I was made renaming the file from .gcode to .txt because of GitHub). Since the "[Svelte example](https://github.com/xyz-tools/gcode-preview-svelte)" loads without any issues, I suspect the problem lies on my end. I have already tried resetting the code via Git to an earlier working state, but the error persists. Currently, I am using the GCode-Preview version "3.0.0-alpha.2," but even the release version 2.18.0 results in the same error. Can you suggest which direction I should take for troubleshooting? To clarify, I have attached the code of my Svelte component.

Thanks, Christian

My svelte component:

```svelte
<script lang="ts">
    import {onMount} from 'svelte';
    import * as GCodePreview from 'gcode-preview';

    import translation from '../../../store/translation';
    import type {GCodeFile} from '../../../types/components';

    import {GCode as GCodeUtils} from '../../../utilities/gcode';

    interface IProperties {
        file?: GCodeFile | null
        invertedZAxis?: boolean
    }

    let {
        file,
        invertedZAxis = false
    }: IProperties = $props();

    let canvas: HTMLCanvasElement;

    //region hooks

    onMount(() => {
        // is dark mode active?
        const html = document.getElementsByTagName('html')[0];
        const bgPreviewDark = '#1e293b';
        const bgPreviewLight = '#d1d5db';

        if (html) {
            // create the observer
            const observer = new MutationObserver((mutations) => {
                for (const mutation of mutations) {
                    // check if the class attribute has changed
                    if (mutation.type === 'attributes' && mutation.attributeName === 'class') {
                        preview.backgroundColor = html.classList.contains('dark') ? bgPreviewDark : bgPreviewLight;
                    }
                }
            });

            // start the observer
            observer.observe(html, {attributes: true, attributeFilter: ['class']});
        }

        // initialize GCode canvas
        const preview = GCodePreview.init({
            canvas,
            lineWidth: 3,
            renderTravel: true,
            renderExtrusion: false,
            backgroundColor: html && html.classList.contains('dark') ? bgPreviewDark : bgPreviewLight
        });

        // everytime the file object changes, the view must be updated
        $effect(() => {
            if (file && file.raw && file.raw.length > 0) {
                preview.clear(); // remove existing items
                preview.processGCode(prepareGCode(file.raw, invertedZAxis)); // draw new code
            }
        });
    });

    //endregion

    /**
     * Prepare the GCode for the preview plugin by removing the line counter and inverting the z axis if necessary.
     *
     * @param lines The lines of gcode as a string-array
     * @param invert If true, all z-axis values are with a minus sign.
     * @return       The modified gcode.
     */
    const prepareGCode = (lines: string[], invert: boolean): string[] => {
        // remove numbers / line counting from the code file
        let prepared: string[] = GCodeUtils.removeLineCounter(lines);

        // change sign of z axis
        return GCodeUtils.setZSign(prepared, invert);
    }

</script>
<canvas bind:this={ canvas } class="rounded-lg w-full"></canvas>

<p class="text-xs font-medium leading-relaxed text-gray-500 dark:text-gray-400/75">
    {#if file}
        { translation.translate('page.file.fusion.preview.gcode.selected.file.title') } { file?.name }

    {:else}
        &nbsp;

    {/if}
</p>

<style>
    canvas {
        cursor: grab;
    }
</style>

```

[1005-Kontur-Holz-6mm-Z2.txt](https://github.com/user-attachments/files/18568787/1005-Kontur-Holz-6mm-Z2.txt)

remcoder:
fixes #285

### DIFF
--- a/src/webgl-preview.ts
+++ b/src/webgl-preview.ts
@@ -116,7 +116,7 @@ export class WebGLPreview {
   /** Width of rendered lines */
   lineWidth?: number;
   /** Height of extruded lines */
-  lineHeight?: number;
+  lineHeight = 0.2;
   /** First layer to render (1-based index) */
   _startLayer?: number;
   /** Last layer to render (1-based index) */
@@ -217,7 +217,7 @@ export class WebGLPreview {
     this.endLayer = opts.endLayer;
     this.startLayer = opts.startLayer;
     this.lineWidth = opts.lineWidth ?? 1;
-    this.lineHeight = opts.lineHeight;
+    this.lineHeight = opts.lineHeight ?? this.lineHeight;
     this.buildVolume = opts.buildVolume && new BuildVolume(opts.buildVolume.x, opts.buildVolume.y, opts.buildVolume.z);
     this.initialCameraPosition = opts.initialCameraPosition ?? this.initialCameraPosition;
     this.renderExtrusion = opts.renderExtrusion ?? this.renderExtrusion;


sophiedeziel:
IIRC, I removed the default value because an undefined lineHeight allowed it to be calculated by the interpreter. That logic is flawed and apparently not tested well enough! Setting it there overrides the one from the job or layer. I have to check the code to confirm if that's the case or not. I think the default should be set elsewhere, but I'm not sure yet where it should be.

remcoder:
Good point. The default is actually both in Path and in ExtrusionGeometry, now that I've checked. I guess I was too hasty with my fix. Btw, the real bug seems to have been introduced when I fixed something else last month. >.<

PR #287, opened 2025-02-17
Support webgl 1

remcoder:
This PR only calls BatchedMesh.addInstance when available. This should hopefully allow threejs r162 to be used, so users can still use browsers that only support webgl 1.

TODO
- [x] test this PR in a webgl 1 browser
- [x] revert the downgrade of threejs. Uses will need to set the versions themselves.

Note:
In Chrome it seems possible to disable webgl2 via the flag `--disable-webgl2`:

`/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --disable-webgl2 "$@"`

These tests confirm this:
- https://get.webgl.org/ (still works)
- https://get.webgl.org/webgl2/ (reports 'While your browser seems to support WebGL2, it is disabled or unavailable.')


remcoder:
I have successfully verified webgl 1 compatibility via the above method. The only requirement is that the consumer builds there own version based on Three r162.

Still, there is no guarantee this will work for future versions of GCode Preview and there is very little we can do to support this type of usage.

That being said, it works! Happy webgl 1 usage! 😄

PR #303, opened 2025-07-07
Update deps

TimTheBig:
This patch updates dependencies, from my testing the demo works fully


TimTheBig:
I will fix the lockfile as soon as I can

remcoder:
hi @TimTheBig, thanks for the PR! 

For the lock file, I think you just need to run `npm ci` locally and commit the resulting lock file. 
When it works again the other checks will be run automatically too: linting, type checks, vitest. You can run them locally too.

Furthermore I see these 2 issues
- live-server was kept at 1.2.1 explicitly due to a security issue
- eslint 9 is not drop-in compatible with 8, my preference would be to stay with 8 for now.

Since deps can be finicky, let's go for the low-hanging fruit. Otherwise it'll just be time sink.

Really appreciate your effort btw!

TimTheBig:
> Furthermore I see these 2 issues
> 
>  * live-server was kept at 1.2.1 explicitly due to a security issue
> * eslint 9 is not drop-in compatible with 8, my preference would be to stay with 8 for now.

Resolved

remcoder:
There are some minor formatting issues left. You can auto-fix them using `npm run prettier:fix`.

remcoder:
I was about to merge but then I did a quick test to see if the resulting build output is different. 
image

Surprisingly the build is 2.5x larger than before which we need to prevent. 

For now I'd suggest we try to find which update causes this and skip that one.

TimTheBig:
> Surprisingly the build is 2.5x larger than before which we need to prevent.

Oh wow that's a lot, I will look into that, some library must have added a big feature.
Does your build system not remove dead code?

remcoder:
Sorry my bad. I accidentally compared it to a dev build while I thought it was a prod build 😅 

So we can merge it! Thanks a bunch for contributing! Much appreciated 😄

TimTheBig:
No problem, thanks for the review and help

PR #309, opened 2025-09-06
Redefine the main class/init to extract many responsibilities out of the renderer

### LINKED ISSUE #232: Redefine the main init function that handles the full lifecycle of the library
Very useful for https://github.com/remcoder/gcode-preview/issues/208
Could be useful for https://github.com/remcoder/gcode-preview/issues/186

Currently, the init function is just a function that creates an instance of the `WebGLPreview` class. Since objects added to a WebGL must be properly disposed of and the different components are split out, a main object should be responsible for managing all the moving parts. 

I started doing a proof of concept: https://github.com/remcoder/gcode-preview/pull/169/files#diff-0e7851300750764bf1798e6ba4dc51f241c919cb43b45190ccfa0b91f30205cc

In the POC above, I only abstracted some interactions so that the lib users never interact directly with the `WebGLPreview` instance. Managing the lifecycle of the app was not started, but it's clearly the main advantage of it.

### LINKED ISSUE #239: Define role and name for the `WebGLPreview` class
The `WebGLPreview` used to be the god class that handled everything. Now that we've extracted a ton of responsibilities out of it, maybe it's time to give it a new name that actually represents what it does (and decide what it should not do).

We are referring to it as the renderer, but it does much more than rendering. There's some code that is pretty easy to extract out because it just does not belong there anymore (drag and drop for example). But it's at the same time a parsing interface, an object builder, an animator, an options manager, and a three.js scene manager.

If we extract out everything but the object building and the three.js scene managing, we're in a very good spot. I would refrain from separating the two too early though because rendering modes will give us plenty of insights on how to get it right. For names, I think it's too early to call it a renderer because of rendering modes. They will probably need a whole new definition of what a renderer is as opposed to what we have right now. `Previewer` sounds right to me. `SceneManager` too. 

I like to ask ChatGPT for naming ideas. Here's what it came up with:

> It sounds like you're looking for a name that reflects the core responsibilities of the class without prematurely locking it into a narrow role like "renderer." Given that it's involved in object building, scene management, parsing, and animation, but you're stripping away certain non-core functions like drag-and-drop, here are a few name suggestions that could fit its role:
> 
> 1. **ScenePreviewer** – Combines both the idea of scene management and previewing, acknowledging the class's wider role without narrowing it down to just rendering.
>    
> 2. **VisualComposer** – Emphasizes its role in assembling and managing 3D objects and scenes, with "composer" hinting at its role in organizing elements.
> 
> 3. **SceneOrchestrator** – A slightly more sophisticated term, implying the coordination of different elements within a scene, including objects, animations, and rendering modes.
> 
> 4. **ObjectManager** – If the focus is more on the management of 3D objects within the scene (after other responsibilities are stripped out), this could work.
> 
> 5. **StageDirector** – Implies a more dynamic role, orchestrating how objects and scenes come together, while allowing room for future functionalities like different rendering modes.
> 
> 6. **SceneAssembler** – Suggests that the class is responsible for assembling and managing the scene components, leaving room for future refinements as more tasks are extracted.
> 
> 7. **Visualizer** – This name is abstract enough to allow for both scene management and rendering without tying it down to any one function too early.
> 
> 8. **SceneArchitect** – If you want to imply that this class is responsible for designing and structuring the scenes, this could be fitting.
> 
> Since rendering modes will impact the definition of a renderer later, these names aim to remain flexible enough to evolve with the project. What do you think of these options?

I kinda like `VisualComposer` and `SceneAssembler`.

sophiedeziel:
Closes https://github.com/xyz-tools/gcode-preview/issues/232
Closes https://github.com/xyz-tools/gcode-preview/issues/239

### Summary

This PR introduces a refactor I wanted to do for a long time! This creates better separation of concerns between G-code processing logic and WebGL rendering operations.

### Key Changes

- Changed the main export `WebGLPreview` → `GCodePreview`, We now have a main class
- Renamed `WebGLPreview` → `SceneManager`, less about the implementation, more about what it does
- Better separation of concerns: `GCodePreview` handles orchestration while `SceneManager` handles all WebGL operations
- The interpreter and the job are now objects of the main class. Only the job is passed to the sceneManager to be rendered
- Extracted some shared types for input
- Several minor cleanups as I saw dead/misplaced lines

### Breaking changes

This introduces a breaking change we'll have to document in the next alpha release, and possibly a migration guide for the next major release. As discussed, let's make the breaking changes now, as the migration is simple. 

To initialize the library, we're going from

```Typescript
import * as GCodePreview from 'gcode-preview';

const preview = GCodePreview.init({
      canvas: document.querySelector('canvas'),
      extrusionColor: 'hotpink'
  });
```  

to

```Typescript

import { GCodePreview } from 'gcode-preview';

const preview = new GCodePreview({
      canvas: document.querySelector('canvas'),
      extrusionColor: 'hotpink'
  });
```

Then, most attributes that used to be on the returned preview object are now accessible using `.sceneManager`. For example, 

```Typescript
preview.backgroundColor = "#ffffff"
```

becomes

```Typescript
preview.sceneManager.backgroundColor = "#ffffff"
```


remcoder:
This is so awesome! I agree this is the cap stone on the big refactor. Well done! 🫶

src/renderer.ts
```diff
@@ -839,7 +714,8 @@ export class WebGLPreview {
   }
 
   private renderBoundingBox(): void {
-    if (!this.job || !this.job.boundingBox.isValid) {
+    if (!this.job) return;
+    if (!this.job.boundingBox.isValid) {
```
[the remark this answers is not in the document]
sophiedeziel: I made that change to not console.log about an invalid bounding box if there is no job at all

demo/js/app.js
```diff
@@ -178,39 +177,45 @@ export const app = (window.app = createApp({
...
 
       watchEffect(() => {
-        preview.backgroundColor = settings.value.backgroundColor;
-
-        if (preview.buildVolume && settings.value.drawBuildVolume) {
-          preview.buildVolume.smallGrid = settings.value.buildVolume.smallGrid;
-          preview.buildVolume.x = +settings.value.buildVolume.x;
-          preview.buildVolume.y = +settings.value.buildVolume.y;
-          preview.buildVolume.z = +settings.value.buildVolume.z;
+        if (!preview) return;
```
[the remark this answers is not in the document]
sophiedeziel: They do

src/extra/dom-utils.ts
```diff
@@ -1,10 +1,10 @@
-import { WebGLPreview } from '../webgl-preview.js';
+import { GCodePreview } from '../gcode-preview.js';
 
 /**
  * Enables drag and drop handling for G-code files
  */
-export function makeDroppable(previewInstance: WebGLPreview): void {
-  const element = previewInstance.canvas;
+export function makeDroppable(previewInstance: GCodePreview): void {
+  const element = previewInstance.renderer.canvas;
```
remcoder: .renderer is a private member so we cannot access it this way.
It should be passed as an arg to makeDroppable, just like in devGui.

remcoder:
Here are some observations but it's just nitpicking really:
- I would vote for having types to go with their impl
- naming suggestion: animateCallback -> onRender or onFrame or onTick or onUpdate or onAnimate
- naming: we now have a renderer that has a renderer.. maybe SceneManager would be better for the outer one?

src/extra/dom-utils.ts
sophiedeziel: This is a point for the renaming of renderer!

This specific renderer is public, the other one is private. The renaming to `SceneManager` would have avoided that confusion

src/renderer.ts
```diff
@@ -613,8 +613,8 @@ export class Renderer {
     this.animationFrameId = requestAnimationFrame(() => this.animate());
     this.controls.update();
     this.renderer.render(this.scene, this.camera);
-    if (this.animateCallback) {
-      this.animateCallback();
+    if (this.onFrame) {
```
remcoder: how about?
`this.onFrame?.()`

sophiedeziel:
I've applied the feedback, and updated the PR description to reflect the new state of the changes!

remcoder (approved):
LGTM!

PR #310, opened 2025-09-08
Manage the preview's rerenders using events

### LINKED ISSUE #236: Preview should handle updates itself
We can remove the necessity for the consumer to call `render()` after updating public props. This way the lib can decide what type of update is needed.

## Outline
Whenever a prop is changed, we queue the change.
Every tick, the queue is read and processed and emptied.

## Update types
 - scene changes can be done immediately w/o requiring the model to be rerendered. Examples:
   - [x] build volume changes
   - [x] clipping planes
   - [ ] ? toggling groups on and off ?
 
- incremental parsing
  - a new chunk of gcode is parsed and translated into a geometry
  - only the new chunk is added to the scene
 
- full render of the model
  - the model is removed from the scene
  - the parsed gcode is rendered anew and added to the scene
  - example: extrusion width was changed

sophiedeziel:
Closes https://github.com/xyz-tools/gcode-preview/issues/236

### Summary 

In order to have the library perform the renders whenever needed, and without the client's intervention (✨ magical ✨), I removed the public `.render` function from the main class, the calls from the demo app, and worked from there to have re-renders performed automatically on any setting change.

I had trouble assigning the default and initial values without triggering 1000 initial renders. I realized that it would work really well with a debounce and an event listener that only starts listening after the initial setup. 

So I introduced a whole events dispatcher that clients can use!

Example:

```Typescript
  const preview = new GCodePreview({
      canvas: document.querySelector('canvas'),
      extrusionColor: 'hotpink'
  });

  preview.addEventListener('animationComplete', () => {
    console.log('Animation complete!')
  });
```

I did not add the ability to remove a listener yet, I can do it separately or as part of this. 

### Key changes

- Added `.addEventListener` that can listen to many different events from the lib. 
  - Converted some of the existing code and logic to use them when appropriate
- I had to convert a ton of properties in `SceneManager` from simple propertied to `_` prefixed with getters/setters. It makes the class very long and verbose, but at least they're all consistent!
- Changed some of the update logic in the demo. Instead of recreating a new preview object each time, I'm reseting it. This allows for an easier management of listeners:
  - Changed the presets structure a bit. Instead of a `init` preset, we can see them as presets for the UI, that is then applied to the preview.
  - Load presets -> merge with defaults -> apply to the demo UI -> UI listeners apply them to the preview
  - The flow above ensure that we push everything to the preview. When the preview can "tell us" something we should update (like the layer count), it's event listeners that we should use. 

### WIP

- The changes to the app broke the initial camera position preset. 
- Tests


src/scene-manager.ts
```diff
@@ -257,6 +281,9 @@ export class SceneManager {
 
     this.initScene();
     this.animate();
+    this.eventsDispatcher.addEventListener(SceneManagerEvent.ANIMATION_COMPLETE, () => {
+      this.setRerenderListeners();
+    });
```
sophiedeziel, line 286: That's the key to re-render automatically on changes. This is setting listeners on all properties that we want to see reflected right away on the scene.

src/scene-manager.ts
```diff
@@ -937,6 +1041,30 @@ export class SceneManager {
...
+      SceneManagerEvent.LAST_SEGMENT_COLOR_CHANGE,
+      SceneManagerEvent.RENDER_TRAVEL_CHANGE
+    ];
+    const rerenderDebounce = 100;
+    this.eventsDispatcher.addEventListener(eventsRequiringRerender, () => {
+      if (this._renderTimeout) clearTimeout(this._renderTimeout);
+      this._renderTimeout = setTimeout(() => {
+        this.render();
+      }, rerenderDebounce);
+    });
```
sophiedeziel, line 1065: debounced event listeners to re-render! ✨ magical ✨

src/.eslintrc.js
```diff
@@ -4,5 +4,14 @@ module.exports = {
   ignorePatterns: ['three-line2'],
   env: {
     browser: true
+  },
+  rules: {
+    'no-unused-vars': [
+      'error',
+      {
+        argsIgnorePattern: '^_',
+        ignoreRestSiblings: true
```
sophiedeziel, line 13: That's how I got rid of the es-lint disable comments when defining a class properties without using them in the constructor. Typescript implicitly uses Rest to assign the values, if I understood correctly.

PR #311, opened 2025-09-08
Replace the rerender event bus with direct ObjectsManager calls

### LINKED ISSUE #278: Travel lines are clipped
Often there are travel moves above the topmost extrusion layer. Due to the clipping planes, of which the max is initialized at the topmost extrusion layer, those travel lines are clipped. 

```
      x--------x  Travel move (INVISIBLE due to clipping)
----------------  clipping plane
================  Layer n (topmost extrusion layer)
================  Layer n-1
================  Layer n-2

```

image

image

### LINKED ISSUE #282: Toggle travel lines w/o rerendering the extrusion
When the extrusion is already fully rendered and then the travel lines are turned on or off, this shouldn't trigger en full render. 

HOW
Extrusion and travel should be created in separate scene groups. These groups can then be toggled on and of individually using the .visible property. 

This is good excuse for introducing the SceneManager also.

### LINKED ISSUE #375: SceneManager never disposes its geometries: every render leaks a full model
## Summary

`SceneManager` never releases the geometries, materials and meshes it creates. Every `render()` pushes a full model's worth of GPU-bound objects onto the private `disposables` array, and the only thing that ever drains it is `dispose()` — tearing down the whole `SceneManager`. `initScene()` removes the previous model from the scene graph but keeps it alive in the array, so each re-render retains another copy.

The drain that would fix this is present but commented out, with no reason recorded in the code (see the comment below for what is probably the real reason — the tube materials are globally cached and shared, so disposing them is unsafe as things stand):

https://github.com/xyz-tools/gcode-preview/blob/develop/src/scene-manager.ts#L690-L693

```ts
// while (this.disposables.length > 0) {
//   const disposable = this.disposables.pop();
//   if (disposable) disposable.dispose();
// }
```

This is a strong-reference leak, not a "did the GPU reclaim it" subtlety: `disposables` is a live field, so the JS objects cannot be collected, so their `WebGLBuffer` wrappers cannot be either. The allocations stay until the `SceneManager` goes away or the context is lost.

## Measurements

Summing the byte length of every attribute array held by `disposables` after each successive `render()` of `demo/gcodes/3DBenchy.gcode` (3511 extrusion paths, 240 layers, 1 tool):

| renders | line mode | tube mode |
|---|---|---|
| 1 | 4.6 MB, 5 entries | 82.4 MB, 3,516 entries |
| 2 | 9.2 MB, 8 entries | 164.7 MB, 7,030 entries |
| 3 | 13.8 MB, 11 entries | 247.1 MB, 10,544 entries |

Growth is exactly linear, and `dispose()` drains it to 0.

Tube mode is ~18x worse per render because it creates one geometry per path plus a `BatchedMesh` sized to the total vertex count, where line mode creates one packed geometry per tool.

This bites in ordinary use: the demo re-renders from a `watchEffect` covering travel color, line width, extrusion width, the tube toggle and the highlight colors. In tube mode each nudge of the extrusion-width input costs ~82 MB that is never given back — a dozen tweaks is ~1 GB, on a model that is about as small as real test cases get.

## The build volume re-push

Restoring the drain is not a one-line change, because `disposables` also holds objects that are still live.

`initScene()` pushes `this._buildVolume` onto `disposables` on **every** render:

https://github.com/xyz-tools/gcode-preview/blob/develop/src/scene-manager.ts#L695-L698

It is also pushed once in the constructor and once per `buildVolume` setter call. So after N renders the same `BuildVolume` instance sits in the array N+1 times, and `dispose()` calls `BuildVolume.dispose()` on it repeatedly. That happens to be harmless today only because `BuildVolume.dispose()` null-checks `_group` and is therefore idempotent.

For a drain, the consequences to work through:

- **`BuildVolume` is long-lived and shared**, unlike the per-render model objects. Disposing it drops its group from the scene. It survives a drain placed at the top of `initScene()` only because `initScene()` re-pushes and calls `update()` immediately afterwards, which rebuilds the group. That is load-bearing ordering that nothing currently documents or tests.
- **The bounding box mesh** is pushed at `renderBoundingBox()` and lives across renders in the same way.
- **Duplicate entries** mean a drain must tolerate disposing the same object more than once, or the array should stop accepting duplicates in the first place.

The underlying problem is that one array mixes two lifetimes: per-render model objects that should die with each render, and session-lived scene furniture (build volume, bounding box) that should not. Splitting those, rather than un-commenting the loop, is probably the actual fix.

## Still to measure

The numbers above are CPU-side attribute arrays measured under a mocked `WebGLRenderer`, which is a reasonable proxy for the vertex data being kept alive but is **not** a GPU measurement. Before settling on a fix we should confirm the real cost:

- `renderer.info.memory.geometries` / `.textures` across successive renders in the real demo, which reads the renderer's own accounting rather than ours.
- Actual VRAM growth in the browser (Chrome task manager GPU memory column, or `WEBGL_debug_renderer_info` plus OS-level GPU monitoring) over a realistic demo session.
- Whether the browser reclaims anything once a drain is in place, i.e. that disposing actually returns the memory rather than just dropping our references.
- A large model as well as the Benchy, since per-path geometry count is what drives tube mode.

That measurement matters for prioritisation: if VRAM tracks the numbers above, this is a crash-in-normal-use bug rather than untidiness.

## Notes

Found while working on #356 (the no-op rendering options). Not fixed there — that PR deliberately stays in scope.

While confirming the above, one adjacent inconsistency turned up in the same area: `materials` has two writers with incompatible indexing. The `extrusionColor` setter writes `this.materials[index]` by *color* index, while `renderPathsAsTubes` `push`es by render order, and `renderPathsAsLines` never registers a material at all — so in line mode the array is empty after a render and the setter builds shader materials nothing draws with. Worth a separate issue if it is not already known.

sophiedeziel:
Fixes #282, fixes #375 and fixes #278. **Alternative to #310**

#310 routes property changes through an event bus. This POC takes the opposite direction and removes `EventsDispatcher` altogether — every notification is a typed callback.

It also delivers what #236 asks for: consumers no longer call `render()` after changing a property — the library decides per property what kind of update is needed.

---

### The two approaches

```mermaid
flowchart TB
  subgraph B["PR 311 — the manager decides, no bus"]
    direction TB
    b1["setter"] --> b2["ObjectsManager"]
    b2 --> b3["group.visible = false"]
    b2 --> b4["uniforms.uColor.value = c"]
    b2 --> b5["material.clippingPlanes = p"]
    b2 --> b6["debounce -> reset -> redraw"]
    b7["notifications"] --> b8["onFrameRendered / onJobUpdated<br/>typed callbacks"]
  end

  subgraph A["PR 310 — every change is an event"]
    direction TB
    a1["setter"] -->|"emit('extrusionColorChange')"| a2(["EventsDispatcher"])
    a2 --> a3["debounce 100ms"]
    a3 --> a4["render()"]
    a4 --> a5["resetScene()"]
    a5 --> a6["rebuild every geometry"]
    a6 --> a7["7023 paths re-walked"]
  end
```

Of `SceneManagerEvent`'s 23 members, 21 were `…Change` events with no listener outside the library — they existed only to trigger a full rebuild. The remaining two were real notifications, and they became callbacks, so the enum and the dispatcher are both gone:

| was | is now |
| --- | --- |
| `frameRendered` | `SceneManager.onFrameRendered` |
| `jobUpdated` | `GCodePreview.onJobUpdated` |
| `streamReadEnd` | `GCodePreview.onStreamEnd` |
| `animationComplete` | dropped — `renderAnimated()` already resolves then |

(The `addEventListener`/`EventName` API only ever existed inside this branch — it was introduced and removed here, so nothing shipped from `develop` loses it.)

**Breaking changes vs `develop`:**

- `GCodePreview.render()` is removed — rendering is reactive; use `preview.sceneManager.render()` / `renderAnimated()` for a manual draw
- `SceneManager`'s constructor narrows to `(opts, job)`; the `onFrame` argument becomes the `onFrameRendered` property
- `SceneManagerOptions.nonTravelMoves` and the `nonTravelmoves` field are removed
- `SceneManager.inches` and `initialCameraPosition` are no longer public
- `processGCode` now returns the `renderAnimated()` promise, and `processGCodeStream` resolves only once the animated render completes

New callbacks, each holding a single assignable listener:

```js
preview.onJobUpdated = (job) => updateUI();
preview.onStreamEnd = () => updateUI();
preview.sceneManager.onFrameRendered = () => stats.update();
```

---

### Where each property goes now

```mermaid
flowchart LR
  subgraph REBUILD["needs new buffers"]
    d1["lineWidth"] --> m4["reset + redraw"]
    d2["lineHeight"] --> m4
    d3["extrusionWidth"] --> m4
    d4["renderTubes"] --> m4
  end

  subgraph CHEAP["in place"]
    c1["extrusionColor"] --> m2["material uniform"]
    c2["travelColor"] --> m2
    c3["ambientLight"] --> m2
    c4["directionalLight"] --> m2
    c5["brightness"] --> m2
    c6["startLayer"] --> m3["clipping planes"]
    c7["endLayer"] --> m3
    c8["singleLayerMode"] --> m3
  end

  subgraph FREE["free"]
    r1["renderTravel"] --> m1["group.visible"]
    r2["renderExtrusion"] --> m1
  end

  style FREE fill:#1a5c2e,color:#fff
  style CHEAP fill:#7a5c00,color:#fff
  style REBUILD fill:#8b1a1a,color:#fff
```

---

### Toggling travels

```mermaid
sequenceDiagram
  autonumber
  participant U as caller
  participant S as SceneManager
  participant O as ObjectsManager
  participant G as scene graph

  rect rgb(60, 20, 20)
  note over U,G: PR 310
  U->>S: renderTravel = false
  S->>S: emit + debounce
  S->>O: resetScene()
  O->>G: discard everything
  S->>O: rebuild 7023 paths
  end

  rect rgb(20, 50, 30)
  note over U,G: PR 311
  U->>S: renderTravel = false
  S->>O: setTravelsVisible(false)
  O->>G: travelMovesGroup.visible = false
  note over G: geometry kept,<br/>re-enable is free
  end
```

---

### Verified in the browser

3DBenchy, 7023 paths, driven through the real API:

| action | result |
| --- | --- |
| travel toggle on/off/on | same uuid, extrusions untouched |
| `extrusionColor` | same uuid, uniform → `00ff00` |
| `travelColor` | same uuid, colour → `ff00ff` |
| `endLayer = 80` | same uuid, `clipMaxY = 24`, model visibly clipped |
| `renderTubes` | `BatchedMesh` ⇄ `LineSegments2`, rebuilt — as intended |

---

### Bugs this surfaced

- `createClippingPlanes(lineWidth, lineHeight)` was passing dimensions where a layer range belonged
- Every rerender appended an empty `LineSegments2` and leaked its material
- `renderPathsAsTubes` pushed a new material per progressive frame instead of caching one per tool
- `renderPathIndex` indexes `job.paths` but was slicing `job.travels`, so re-enabling travels after a completed render drew nothing
- `clear()` left the old `ObjectsManager` in `disposables`, double-disposing it
- Rebuilt tube materials lost the layer range

And found in review, fixed on this branch:

- The global color-keyed material cache aliased same-color tools and resurrected disposed materials — removed (it was also the blocker #375 identifies for safe disposal)
- `clear()` reset `renderTubes` and the lighting to defaults for the next job
- An unset start layer baked `NaN` into every clipping plane and clip uniform (undefined behavior in GLSL)
- A scalar `extrusionColor` only recolored tool 0
- devMode created two `DevGUI`s: the constructor duplicated the setter's `initGui()`
- `renderAnimated()`'s completion promise was not propagated by `processGCode` / `processGCodeStream`

---

### Numbers

| | before | after |
| --- | --- | --- |
| `SceneManagerEvent` members | 23 | 0 — enum deleted |
| `emit()` calls | 19 | 0 |
| `EventsDispatcher` | 1 class + bus | deleted |
| `scene-manager.ts` (vs `develop`) | 1075 lines | 904 |
| tests (vs `develop`) | 266 | 355 |
| coverage of new code | 23% | 100% |

---

### Still open

- Build volume and bounding box aren't moved into the manager yet — done in #380, stacked on this PR
- `renderPathIndex` still conflates paths / travels / toolPaths in `renderAnimated`
- Progressive rendering still leaves per-frame chunks in the scene (#355) — and with rebuilds now rare, they persist until a dimension property changes
- `topLayerColor`, `lastSegmentColor` and `disableGradient` are still stored but never read (#356)
- ~Travel moves above the top layer are still clipped (#278)~ — fixed here: a layer bound at the very end of the stack no longer produces a clipping plane


remcoder:
In terms of scope:
- let's leave #356 out his. Low prio for me
- I'd vote to leave #355 out too for now.. to keep things simple. let's stabilze first

remcoder:
@sophiedeziel 
btw, let's do some test before merging. I'm thinking off:
- large model test
- perf regression test
- slow computer / mobile test
- the frame components ( react / vue / svelte) should be tested, for lifecycle test
- mem leak test
anything else?
Maybe some are already done?

sophiedeziel:
> perf regression test

I measured yesterday, we actually have a performance gain with fewer re-renders, and some re-renders that start at some point of the flow instead of doing a whole round trip

PR #313, opened 2025-09-12
Instant update narrow line color

remcoder:
Color of narrow extrusion/travel lines is updated instantly. 

The way I access the travel lines and extrusion lines in the scene:
- .userData a reserved object on 3d objects (meshes, groups) with a custom prop for 'lineType'.
- a function to 'query' the scene by traversing the scene graph and filtering on that lineType


remcoder:
Parking this PR for now. The instant color update works and is a real improvement, but the approach overlaps with the object-manager direction in #310/#311. Those PRs track meshes explicitly (O(k) direct access) instead of discovering them by scene traversal per color set (O(n) full walk), and they're the intended architecture for exactly this kind of update. If that work lands, this traversal-based code gets refactored away.

Issue #179, opened 2024-06-28
Implement missing gcodes that affect position


remcoder:
done in v3.0:

- [x] G20 - Set units to inches
- [x] G21 - Set units to mm
- [x] G28 [X Y Z] - Home axis

planned for v3.1+
- [ ] G90 - Absolute positioning mode
- [ ] G91 - Relative positioning mode
- [ ] G92 [X Y Z E] - Set current position
- [ ] M206 [P T X Y Z] - Set home offsets
- [ ] M428 - Home Offsets Here
- [ ] absolute E distances

sophiedeziel:
My parser refactor is not so unnecessary it seems! I'll try to split that in a new branch. 

I have some state management that will be useful.

remcoder:
There's a new architecture with a revamped parser in v3.0 so let't move this to v3.1+, apart from the commands that are already in v3.0.

dansity:
Hi,
Seems like G3 for travel is missing also? Or is it simplified?

<img width="1125" height="913" alt="Image" src="https://github.com/user-attachments/assets/48c45701-34d7-4417-b60a-128c7a7d2667" />

For reference, same gcode in ncviewer:

<img width="1326" height="801" alt="Image" src="https://github.com/user-attachments/assets/648b4779-aff2-43d2-90d1-5195ae6f3d7b" />

remcoder:
Could be a bug with full circles (just a wild guess). 

G2/G3 are generally working so this seems like an edge case.

dansity:
ok, i will open a separate issue. Its reproducible

PR #345, opened 2026-08-20
feat: support G90 and G91 positioning modes

### LINKED ISSUE #432: Support G90/G91 (absolute/relative positioning)
Interpret X/Y/Z coordinates as absolute or relative depending on mode.

**Seen in:** every file's start gcode sets G90; G91 typically wraps Z-hops and end-gcode lifts.
**Behavior:** mode flag for XYZ interpretation; in relative mode each move adds to the current position.
**Gotchas:** in Marlin/Prusa, G90/G91 also affects E unless M82/M83 overrides it; on RepRapFirmware they never affect E.
**Related:** #419, #433

Docs: [G90](https://marlinfw.org/docs/gcode/G090.html) / [G91](https://marlinfw.org/docs/gcode/G091.html)

PozziTiv4ik:
Closes #432. Part of #179.

G90/G91 now select absolute/relative positioning for linear moves, arc endpoints, and probes. Previously these commands were ignored on `develop`. For example, `G0 X10 / G91 / G1 X2` now ends at X12, while switching back with `G90 / G1 X2` ends at X2 when there is no workspace shift.

## Dialect and extrusion

This follows Marlin's [G90](https://marlinfw.org/docs/gcode/G090.html), [G91](https://marlinfw.org/docs/gcode/G091.html), [M82](https://marlinfw.org/docs/gcode/M082.html), and [M83](https://marlinfw.org/docs/gcode/M083.html) ordering. G90/G91 select the mode for XYZ **and E**. A subsequent M82/M83 overrides E alone; a later G90/G91 clears that override. Firmware dialects with permanently independent E modes are not inferred automatically.

This is a deliberate rendering and statistics change for files that relied on G90/G91 being ignored, or used G90/G91 after M82/M83. `G91` followed by three `G1 X10 E0.5` moves now deposits 1.5 mm of filament across all three segments, rather than rendering the last two as travel. Mode changes keep the tracked extruder position intact, and compose with G92 E resets, retractions, unit changes, and incremental parsing.

## Coordinate resolution and probes

Coordinates and offsets are normalized to millimeters before state mutation. The handlers choose between two separate state operations: absolute motion uses `physical = logical + positionShift`, and relative motion uses `physical = (currentPhysical ?? 0) + offset`. Applying the G92 shift to an offset would count it twice. I/J/R remain arc offsets in either mode, and omitted axes retain their previous values, including `undefined`.

G31 and G38.2–G38.5 now use the same coordinate resolution. The preview's assumed Z0 probe trigger is applied **after** resolving the physical target: `G28 / G0 Z20 / G91 / G38.2 Z-10` reaches Z10, not Z0. A subsequent downward probe crossing the known origin plane stops at Z0 as before.

A relative move from an unknown Z uses an assumed zero for rendering without establishing the probe's origin. For example, `G91 / G1 Z2 / G90 / G38.2 Z-5` reaches Z-5 instead of inventing contact at Z0. The state tracks that provenance rather than requiring `isHomed` for every probe: existing CNC touch-off files establish a preview coordinate with an absolute Z move without issuing G28, and must retain their converging probe cycles. Homing or an explicit absolute Z target establishes the origin; G92 only relabels it.

## Arc behavior and compatibility

Absolute endpoint-less I/J arcs draw a full circle, as do their relative-mode counterparts. Their regression tests now live with the arc handler rather than in positioning-mode tests, and assert the complete signed revolution in both directions. This absolute-mode behavior is already present in the current integration base, `021bb8060ecdd1b37b2e2a6c53d34817b6997b22`; a direct baseline comparison confirms identical output for `G0 X10 Y20 / G2 I5 J0`. Absolute Z0 and monotonic helical interpolation regression coverage is retained.

The branch includes current `develop`, preserving its millimeter normalization, filament-consumption accounting, and Node 24 / Rolldown build. All seven shipped `demo/gcodes/*.gcode` files produce identical path classifications, coordinates, dimensions, and filament totals against that base. The comparison hashes complete path contents, rather than relying only on vertex sums. This includes the `mach3.gcode` touch-off example.

## State API

`State.positioning` stores the modal XYZ choice. `State.moveTo` and `State.moveBy` provide the separate coordinate operations requested in review without teaching the state to choose a mode. `State.hasKnownZ` exposes the provenance needed by the probe handler; its assumed-origin bookkeeping stays private. The command barrel adds `setAbsolutePositioning` / `setRelativePositioning` for dispatch. No existing symbol is renamed or removed. README documents the supported commands and the selected extrusion convention.

## Validation

Validated on Windows with Node **24.21.0**, matching `.nvmrc`, and three.js **0.185.1**:

- Full coverage run: **1030 tests pass in 45 files**, no type errors; every source file meets 100% statement, branch, function, and line coverage. No coverage exclusions or thresholds changed.
- The new regression cases produced **13 failures before the fixes** and pass afterward.
- `npm run typeCheck`, `npm run lint`, `npm run build`, `npm run typedoc`, and `git diff --check` pass.
- Packaging smoke passes: clean consumer installation, runtime exports, shared JavaScript/declaration chunks, and all four TypeScript resolution modes. A local uncommitted Windows launcher adapts the packaging script's `execFileSync('npm', ...)` to Node plus the npm CLI; the test assertions and repository scripts are unchanged.
- Parser-to-interpreter tests cover mode ordering, shifted linear/arc/probe motion, unknown and omitted axes, partial chunks, inch/mm transitions, and extrusion accounting. Positioning pipeline tests are consolidated in `src/__tests__/interpreter.ts`.
- All seven demo fixture comparisons pass against the current integration base.

## Upstream checks

The updated head is `65f20dc283913919dee061a77681d7ebc066b756`, with no merge conflict against `develop`. The fork's new [unit-test run](https://github.com/xyz-tools/gcode-preview/actions/runs/35627080850) and [Firebase workflow run](https://github.com/xyz-tools/gcode-preview/actions/runs/35627080867) are both `action_required`: they need approval from an upstream maintainer before proceeding. The local results above are complete; these upstream runs have not passed yet. The Firebase job retains the repository's existing same-repository-only condition, so a fork PR does not publish a preview.


src/state.ts
```diff
@@ -5,25 +5,25 @@
...
+  y: number = 0;
   /** Current Z position */
-  z: number;
+  z: number = 0;
   /** Current extrusion amount */
-  e: number;
+  e: number = 0;
   /** Currently active tool */
-  tool: number;
+  tool: number = 0;
```
sophiedeziel: There is a semantic difference between `undefined` and 0. Those should not be set by default.

remcoder: right, good catch

remcoder: hmm, printers actually init e=0, tool=0, units=mm and positioning=absolute.

for x,y,z you need to home first. But we currently have no homed true/false flag. (woops)

unless you want to allow someone instancing new State with undefineds, there is now difference in behavior by this commit.

In fact, I'd argue that until we implement a homed flag, it is best to init to proper values to prevent NaN's later...

remcoder: and I'd argue we might want to remove the 'initial' static initializer as it adds no value then.

remcoder: Another option could be to leave the State object dumb. The interpreter should then make the choice. Which is nice bc in theory there could be a different interpreter simulating a different firmware. In the case of different interpreters we'd want objects like State to be dumb.

remcoder: But for now I'd opt for the simplest solution that doesn't break behavior: Remove 'initial' and init props in the class.. Agree?

remcoder: hmm actually, let's just roll this back and discus this later.. Them rabbit holes... :-)

src/interpreter/commands/resolve-position.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @remarks
+ * An omitted axis keeps its current position. In relative mode the value is an
+ * offset from the current position; in absolute mode it is the target itself.
+ */
+export function resolvePosition(value: number | undefined, current: number, relative: boolean): number {
+  if (value === undefined) {
+    return current;
+  }
+  return relative ? current + value : value;
+}
```
sophiedeziel, line 16: `current`'s type should be `number | undefined` if we look at the state's coordinate properties types. It is undefined when it's not homed. 

In the implementation, `undefined` should be handled gracefully for `current` too. I think that coercing it to 0 is good enough and would allow starting a render even if in theory, the printer does not actually know where the printhead is. Virtual printers don't risk crashing in the printbed!

src/interpreter/commands/resolve-position.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Resolves a move parameter to a target coordinate
+ * @param value - Coordinate from the command, or `undefined` when the axis is omitted
+ * @param current - Current position of the axis
+ * @param relative - Whether relative positioning (G91) is active
+ * @returns The target coordinate for the axis
+ * @remarks
+ * An omitted axis keeps its current position. In relative mode the value is an
+ * offset from the current position; in absolute mode it is the target itself.
+ */
+export function resolvePosition(value: number | undefined, current: number, relative: boolean): number {
```
sophiedeziel, line 11: I kinda like the idea of extracting a function, but I wonder where is the ideal home for it. Probably the `State` should expose two methods to move to/by with x,y,z, and the interpreter uses the relative mode to decide which method to call on the state. I don't think the state should be smart to the point of having logic to handle two modes in the same method. That's the command and interpreter's role to know when and how to change the state.

src/interpreter/commands/resolve-position.ts
PozziTiv4ik: Addressed by replacing the helper with `State.moveTo` / `State.moveBy`. Relative moves now use `(current ?? 0) + offset` for each supplied axis, while omitted axes remain unchanged (including `undefined`) and `isHomed` stays false. Parser-to-interpreter tests cover both unknown and omitted axes through G0, G1, G2, and G3. The full suite passes on Node 22.12.0: 960 tests and 100% coverage, plus typecheck, lint, and build.

src/interpreter/commands/resolve-position.ts
PozziTiv4ik: Implemented the two-method approach: `State.moveTo(x, y, z)` handles absolute logical targets and `State.moveBy(x, y, z)` handles physical offsets. The G0/G1 and G2/G3 handlers choose between them using the active positioning mode; neither state method switches modes. This also keeps G92 shifts out of relative offsets, while preserving them for absolute targets. Added parser-to-interpreter coverage for G90/G91 switching with G92/G92.1, shifted I/J and R arcs, full circles, and chunk boundaries.

sophiedeziel (dismissed):
This PR seem to be in a pretty good state now!

I ran out of tokens for the current session, so I could not run `claude /lessons-learned-review` I created earlier.

src/interpreter/commands/set-positioning.ts
```diff
@@ -0,0 +1,20 @@
+import type { CommandHandler } from '../../interpreter';
+
+/**
+ * Executes a G90 command to set motion commands to use absolute coordinates
+ * @param command - GCodeCommand containing the command
+ * @param job - Job instance to update
+ */
+export const setAbsolutePositioning: CommandHandler = (command, job) => {
```
sophiedeziel, line 8: **[major] G91 leaves E absolute even when the file never emitted M82/M83**

**Problem** — these handlers set only `state.positioning`. In Marlin/Prusa/Klipper/RRF, `G91` sets the relative bit for **all** axes including E; `M82`/`M83` only override E afterwards. Issue #432 names this exact gotcha: "also affects E **unless** M82/M83 overrides it".

**Example** — measured on this head: `G28 / G91 / G1 X10 E0.5` x3 gives `stats.extrusionDistance` = **0.5** and paths `Extrusion, Travel` — the second and third segments render as travel. Marlin extrudes 1.5 mm across three extruding segments. Slicer output is safe (all 7 demo fixtures emit `G90` *before* `M83`), so the exposure is hand-written and CNC-adjacent G-code.

**Recommendation** — the description discloses this as deliberate scope, which is a fair call. But #432 asks for the override form specifically, so either track "has an explicit M82/M83 been seen" and let G90/G91 drive E until then, or file the follow-up and link it here so it isn't lost when #432 closes.

src/state.ts
```diff
@@ -72,6 +74,20 @@ export class State {
...
 
+  /** Moves to logical coordinates, applying the G92 shift only to supplied axes. */
+  moveTo(x: number | undefined, y: number | undefined, z: number | undefined): void {
+    if (x !== undefined) this.x = x + this.positionShift.x;
+    if (y !== undefined) this.y = y + this.positionShift.y;
+    if (z !== undefined) this.z = z + this.positionShift.z;
+  }
+
+  /** Moves by physical offsets, assuming zero only for supplied, unknown axes. */
+  moveBy(x: number | undefined, y: number | undefined, z: number | undefined): void {
```
sophiedeziel, line 85: **[minor] A relative move on an un-homed printer silently makes the axis "known", defeating probe's un-homed guard**

**Problem** — `moveBy` assumes 0 for a supplied unknown axis, so `state.z` becomes a number while `isHomed` stays `false`. `probe.ts:44` keys its "un-homed Z, real position unknown" branch off `state.z !== undefined`, not `isHomed`.

**Example** — measured on this head: `G38.2 Z-5` alone ends at Z**-5** (the documented un-homed behavior). `G91 / G1 Z2 / G90 / G38.2 Z-5` ends at Z**0**, because the relative move populated Z from an assumed origin and the probe now takes the known-Z clamp branch. The Z is no more known than it was before.

**Recommendation** — switch probe's guard to `state.isHomed` (which is what its doc comment actually describes), or have `moveBy` record that the axis is assumed rather than known.

Minor aside on the same block: `moveTo`/`moveBy` use one-line doc comments, while every other member of `state.ts` carries full `@param`/`@returns` tags.

src/interpreter/commands/arc-move.ts
```diff
@@ -42,12 +35,19 @@ export const makeArcMove = (options: ArcTessellatorOptions = {}): CommandHandler
...
+    } else {
+      state.moveTo(x, y, z);
+    }
+
+    // The tessellator emits the exact endpoint; omitted axes retain their
+    // previous (possibly unknown) coordinates in the state.
     arcTessellator.tessellate(
       from,
-      { cw, x, y, z, i, j, r },
+      { cw, x: state.x, y: state.y, z: state.z, i, j, r },
```
sophiedeziel, line 50: **[process] Undisclosed behavior change: endpoint-less arcs now draw a full circle in absolute mode too**

**Problem** — passing `state.x/y` (post-move) instead of `command.params` means an omitted axis reaches the tessellator as a *number equal to the start*, which flips `wholeCircle` on. That's correct per Marlin/RRF — but it is not a G91-only change.

**Example** — measured `G0 X10 Y20 / G2 I5 J0` with no G91 anywhere: develop emits **9** vertices (a degenerate point), this head emits **75** (a real circle). The description says the full-circle fix is "now also present upstream" — it isn't; develop still degenerates, so this PR introduces it. Impact on shipped content is nil: 0 of 11,045 arcs across the 7 demo fixtures omit X/Y.

**Recommendation** — move the full-circle line out of the "fixes already upstream" paragraph and into "Commands and behavior" as an absolute-mode change, and move its covering test out of the positioning-mode block, since it has nothing to do with G90/G91.

src/__tests__/interpreter/positioning-modes.ts
```diff
@@ -0,0 +1,91 @@
+import { Interpreter } from '../../interpreter';
+import { Job } from '../../job';
+import { Parser } from '../../parser/gcode-parser';
+
+const run = (lines: string[]) => new Interpreter().execute(new Parser().parseGCode(lines).commands);
+const vertices = (job: Job) => job.paths.flatMap((path) => path.vertices);
+
+describe('positioning modes with unknown axes and workspace shifts', () => {
```
sophiedeziel, line 8: **[process] The same subject has two test homes**

**Problem** — `src/__tests__/interpreter.ts:126` adds a `describe('positioning modes through the whole pipeline')` block, and this file adds a second parser-to-interpreter block for the same feature. The rest of `src/__tests__/interpreter/` mirrors source files 1:1 (`commands/arc-move.ts` <-> `src/interpreter/commands/arc-move.ts`); there is no `src/interpreter/positioning-modes.ts` for this one to mirror.

**Recommendation** — fold this file into the existing block in `src/__tests__/interpreter.ts`.

sophiedeziel (commented):
## Lessons-learned review — head `92a211f`

Ran the repo's `lessons-learned-review` checklists against `origin/develop` (`7e94ce8`). Sorry for the whiplash on the approval — I'd queued this review before approving and dismissed the approval so the findings below get a look first. Nothing here is structural; the core model is right.

**Verification on head:** `vitest run` 960 pass / 46 files · `--coverage` 100% statements, branches, functions, lines (1793/1793) · `tsc --noEmit` clean · prettier + eslint clean · `typedoc` no warnings. All four validation claims in the description hold.

**Regression check on real files:** parsed all 7 `demo/gcodes/*.gcode` on this head and on `develop`, comparing path count, vertex count and vertex sum. **Byte-identical on every fixture** (3DBenchy: 7023 paths / 328446 verts / sum 22022862.6941 on both). No fixture contains `G91`, and 0 of 11,045 arcs omit X/Y, so nothing in the demo set moves.

**Mutation checks:** adding `positionShift` inside `moveBy` turns 3 tests red; reverting the arc `tessellate` call to raw `command.params` turns 5 tests red across 3 files. The new tests assert behavior, not just execution.

What's good: `moveTo` applies the G92 shift and `moveBy` never does, I/J/R stay offsets, `from` is captured before the state moves, and every guard uses `!== undefined` / `??` rather than truthiness, so `X0`/`Y0`/`Z0` targets survive. The findings are about handlers and docs the change didn't follow through to.

Four inline comments below. Three more can't be anchored to the diff:

---

### [major] G38/G31 probe moves ignore the new positioning mode

**Problem** — `src/interpreter/commands/probe.ts:33-35`: `probe` is a move handler that writes `state.x/y/z`, but it wasn't migrated to `moveTo`/`moveBy` and still treats every target as absolute-plus-shift.

**Example** — measured on this head: `G28 / G0 X10 Y10 Z20 / G91 / G38.2 Z-10` ends at **Z0**, not Z10, while `G91 / G1 Z-10` in the same file correctly lands on Z10. `G91 G38.2 Z-10 F100` is the canonical touch-plate sequence, so this is the common shape — and the negative-Z clamp makes it plunge to the origin plane rather than merely stopping at the wrong height.

**Recommendation** — route probe through the same branch (`state.moveBy` / `state.moveTo`), keeping the Z-trigger clamp applied to the *resolved* physical Z, and add a `G91` case to `src/__tests__/interpreter/commands/probe.ts`.

### [process] README supported-commands table not updated

**Problem** — `README.md:104-114` lists every G-code with a handler (G0/G1, G2/G3, G20/G21, G28, G31, G38.2–G38.5, G92, G92.1, T0–T7). `G90` and `G91` are now registered in `src/interpreter.ts:62-63` but missing from the table, directly above the line "Commands without a handler are parsed but ignored by the interpreter."

**Recommendation** — add `| ` + "`G90` / `G91`" + ` | absolute / relative positioning |`, plus a sentence recording the deliberate XYZ-only scope.

### [process] PR doesn't close its issue

**Problem** — the description references #179 (the umbrella) but not **#432 "Support G90/G91 (absolute/relative positioning)"**, the sub-issue this implements, which is still open.

**Recommendation** — add `Closes #432` to the description.

---

## Latent (no failure today)

**Relative arc with an unknown omitted axis degenerates to a straight line.** `G91 / G2 X10 I5 J0` with Y never homed passes `y: undefined` to the tessellator, so `finalTheta` is NaN, the segment loop is skipped and only the endpoint is emitted. Pinned as intended by `positioning-modes.ts:20`, and masked because the axis genuinely has no known value. It would only become visible if `moveBy`'s assumed origin were ever promoted to a real coordinate. This head is strictly better than develop here: a *known* omitted axis now tessellates correctly where develop produced NaN.

## Out of scope / pre-existing — not for this PR

- **Unit switches corrupt stored coordinates.** `G91 / G1 X10 / G20 / G1 X1` → `state.x = 11`, mixing mm and inches. Absolute mode has the identical bug on develop: neither `moveTo` nor the old `x + positionShift.x` converts by `state.units`.
- **The README table lists no M-codes at all**, though `m82`/`m83` have handlers.
- **G28 doesn't reset `positionShift`**, so homing after a G92 keeps the workspace shift.

Happy to file the first two of these as issues.

## Classes run vs skipped

| Class | Result |
| --- | --- |
| gcode-spec | 3 findings — G38 not mode-aware, E excluded from G91, full-circle semantics |
| behavior-changes | 1 finding — verified byte-identical output on all 7 fixtures, one undisclosed change |
| falsy-zero | nothing found — `moveTo`/`moveBy` use `!== undefined`, `moveBy` uses `?? 0`; X0/Y0/Z0 targets survive |
| nan-guards | nothing found — parser drops non-finite params upstream; the tessellator's non-finite `totalSegments` guard still covers the new call shape |
| test-rigor | 1 finding — placement only; two mutations confirmed the tests bite, coverage is 100% |
| optional-guards | 1 finding — `state.z !== undefined` used as a proxy for `isHomed` in probe |
| streaming-chunks | nothing found — `positioning` lives on `State`, survives chunk boundaries, pinned by `positioning-modes.ts:74` |
| public-api | nothing found — `State.positioning`, `moveTo`, `moveBy` are additive; no renames; typedoc picks up the new command file via the `src/interpreter/commands/*.ts` entry point |
| loop-index | nothing found — no index arithmetic added to source; the only new loop is a test-side angle sum |
| render-allocations | nothing found — `moveTo`/`moveBy` are in-place writes, no per-move allocation added |
| lifecycle-dispose | skipped — no three.js resources, listeners or constructor ordering touched; `positioning` resets with the fresh `Job` in `gcode-preview.ts:160,177` |
| demo-impact | skipped — no options, GUI controls or presets added; no fixture uses G91 |
| workflows | skipped — no `.github/` or `package.json` publish-field changes |

PR #346, opened 2026-08-20
Support three.js r166-r185 with matrix CI coverage

### LINKED ISSUE #308: Update threejs version
Update threejs version maybe with a `>=` requirement for future proofing and better resolving.

remcoder:
Closes #308

## Summary

Relaxes the `three` dependency from a hard pin (`0.178.0`) to a supported range `>=0.166.0 <0.186.0` (r166–r185), and proves that range works by running the full CI suite (build, test, typeCheck, lint) against **every minor three.js version** in the range via a CI matrix.

## Behavior change

This is a dependency-range relaxation only. Rendered output is unchanged; the change makes the library compatible with (and tested against) three.js r166 through r185, including the current r185.

## Why a matrix

Per #308, three.js does not use semver, so each minor can introduce breaking changes. The matrix pins CI to test both the exact min (r166) and max (r185) bounds — and now every minor in between — so a regression at any version surfaces in CI instead of in consumers' projects.

## What was added/changed

- `package.json`: `three` → `>=0.166.0 <0.186.0`; adds `@webgpu/types` as a devDependency
- `.github/workflows/run-tests.yml`: matrix with all 20 minor versions (r166–r185), each paired with its matching `@types/three`
- `tsconfig.json`: added `node_modules/@webgpu` to `typeRoots`
- `src/__tests__/three-version.ts`: guards the declared dependency range and asserts the installed three version stays within it
- `src/__tests__/renderer-smoke.ts`: runtime smoke tests that mock **only** `WebGLRenderer` and construct the real `BatchedMesh`, `LineSegments2`, `LineMaterial`, and `OrbitControls` through the full parse → interpret → render pipeline, then dispose cleanly

## Verification

All 20 matrix combos verified locally: `test`, `typeCheck`, `build`, `lint` pass on each. `npm run check` passes at the lock version (0.178.0).

## Notes

- three.js r166–r185 release notes contain **no breaking changes** to the APIs this library uses (verified against the official migration guide).
- `three@0.180.0` initially failed `typeCheck` due to an upstream bug in `@types/three@0.180.0` (missing `/// <reference types="@webgpu/types" />` in `ExternalTexture.d.ts`); resolved via the `@webgpu/types` typeRoots entry.
- Future note: r186 adds `Object3D.dispose()`; `LineBox.dispose()` should call `super.dispose()` when the max bound is next bumped.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -20,6 +20,52 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
+      matrix:
+        # test against the min and max supported three.js versions (see #308)
+        include:
+          - three-version: 0.166.1
+            types-version: 0.166.0
+          - three-version: 0.167.1
+            types-version: 0.167.2
+          - three-version: 0.168.0
+            types-version: 0.168.0
+          - three-version: 0.169.0
+            types-version: 0.169.0
+          - three-version: 0.170.0
+            types-version: 0.170.0
+          - three-version: 0.171.0
+            types-version: 0.171.0
+          - three-version: 0.172.0
+            types-version: 0.172.0
+          - three-version: 0.173.0
+            types-version: 0.173.0
+          - three-version: 0.174.0
+            types-version: 0.174.0
+          - three-version: 0.175.0
+            types-version: 0.175.0
+          - three-version: 0.176.0
+            types-version: 0.176.0
+          - three-version: 0.177.0
+            types-version: 0.177.0
+          - three-version: 0.178.0
+            types-version: 0.178.1
+          - three-version: 0.179.1
+            types-version: 0.179.0
+          - three-version: 0.180.0
+            types-version: 0.180.0
+          - three-version: 0.181.2
+            types-version: 0.181.0
+          - three-version: 0.182.0
+            types-version: 0.182.0
+          - three-version: 0.183.2
+            types-version: 0.183.1
+          - three-version: 0.184.0
+            types-version: 0.184.1
+          - three-version: 0.185.1
+            types-version: 0.185.4
+
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/run-tests.yml
```diff
@@ -20,13 +20,61 @@ jobs:
...
+          - three-version: 0.181.2
+            types-version: 0.181.0
+          - three-version: 0.182.0
+            types-version: 0.182.0
+          - three-version: 0.183.2
+            types-version: 0.183.1
+          - three-version: 0.184.0
+            types-version: 0.184.1
+          - three-version: 0.185.1
+            types-version: 0.185.4
```
sophiedeziel, line 67: OOOHHHHH <3 <3 <3

package.json
```diff
@@ -61,6 +62,6 @@
   },
   "dependencies": {
     "lil-gui": "^0.20.0",
-    "three": "0.178.0"
+    "three": ">=0.166.0 <0.186.0"
```
sophiedeziel, line 65: That is going to be so much simpler for developers

sophiedeziel, line 65: To bump the upper bound, I recommend creating a skill that any coding agent (or just a human reading skills as documentation) can use when three.js releases a new version. That skill would have instructions on where to find the changelog in order to check the breaking changes, and the list of files to update if we're not affected by breaking changes.

sophiedeziel (approved):

package.json
remcoder: Good point! I don't have any experience in creating agent skills (I don't have the SKILL hah!) so happy to follow your lead there.

sophiedeziel: I can create one! They are essentially documentation. The shape and language does not matter as much as when we were in early days of AI coding agents. Most harness support slash commands to invoke the skill and that's how simple it would be.

PR #348, opened 2026-08-21
Add test coverage tooling

sophiedeziel:
Adds test coverage measurement via `@vitest/coverage-v8`

Brings `gcode-parser.ts`, `extrusion-geometry.ts`, and `scene-manager.ts` to 100% coverage. 

Enforces per-file 100% thresholds in the Vitest config and gates CI on `npm run test:coverage` so coverage can't silently regress.

## Coverage still to do

Overall project coverage sits at **~79% statements / ~71% branches**. Files not yet covered, roughly worst-first:

| File | Stmts | Branch | Funcs | Lines |
|------|------:|-------:|------:|------:|
| `dev-gui.ts` | 0% | 0% | 0% | 0% |
| `extra/dom-utils.ts` | 0% | 0% | 0% | 0% |
| `interpreter.ts` | 47% | 35% | 94% | 47% |
| `build-volume.ts` | 63% | 35% | 71% | 65% |
| `bounding-box.ts` | 67% | 33% | 60% | 67% |
| `helpers/line-box.ts` | 80% | 57% | 75% | 86% |
| `gcode-preview.ts` | 92% | 79% | 88% | 95% |
| `thumbnail.ts` | 92% | 100% | 75% | 92% |
| `helpers/grid.ts` | 93% | 67% | 67% | 96% |
| `path.ts` | 94% | 88% | 80% | 97% |
| `indexers.ts` | 96% | 96% | 100% | 96% |
| `job.ts` | 98% | 84% | 100% | 98% |

As each file reaches 100%, add it to the `thresholds` map in `vitest.config.mts` to lock it in.

### `closed` bug fix

In `extrusion-geometry.ts`, `generateSegment(closed === false ? ...)` read a bare `closed` identifier, which resolved to the global `window.closed` — not `this.parameters.closed`. Since `parameters.closed` was hardcoded to `false`, the closed-tube path was dead code, so the branch is removed rather than parameterized: browser behavior is unchanged, and DOM-less construction no longer crashes on the missing global. The field and its type/JSDoc are deleted, keeping the file honestly at 100% in the thresholds map. The test that toggled the global to cover the dead branch is replaced by a DOM-independence regression test (`delete globalThis.closed`, construct, assert vertex count).

### Regression-shaped scene-manager tests

The scene-manager suite's weak assertions (`not.toThrow()`, bare `toHaveBeenCalled()`, `Object.create(SceneManager.prototype)` stand-ins) were replaced with state assertions against real `SceneManager` instances, enabled by mocking only `three`'s `WebGLRenderer` (the same pattern as the renderer smoke test — cameras, materials, geometry and OrbitControls stay real). Constructor options are now verified on the resulting instance, so the two mutation probes from review (blackening the `travelColor` / `topLayerColor` constructor assignments) each fail the suite, as do clamp-dropping and callback-dropping mutations in the setters. The one remaining non-public-path test (a material without uniforms) injects the malformed state directly and carries an inline justification. Coverage stays at 100% on all axes for the three gated files, and the now-reachable `orthographic ?? false` branch lost its stale `v8 ignore`.

Assisted by Claude Code - Opus 4.8


remcoder (approved):
Coverage tooling looks solid. Verified locally: `npm run test:coverage` passes 196/196 with the per-file 100% thresholds actually enforced.

**Comments (non-blocking):**

1. **Rebase needed / conflict is not cosmetic.** The branch predates the three.js version-matrix CI (#346) and the r185 bump (#357). As it stands, the workflow diff *deletes* the three.js matrix strategy — merging as-is would silently drop range testing (and the package.json pins `three` to `0.178.0`, reverting the `^0.185.4` / range change). Please rebase onto develop and keep the matrix, changing only `npm run test` → `npm run test:coverage`.

2. **The `closed` latent bug is worse than documented.** In the test env `globalThis.closed` is `undefined` (so `closed === false` is `false` → takes the `0` branch), but in a browser `window.closed` is `false` (takes `points.length - 1`). So the *browser* code path is the opposite of what the test exercises. The test correctly documents it, but this deserves a proper fix (`generateSegment(this.parameters.closed ? 0 : points.length - 1)`) — happy to take it as a follow-up so it doesn't block this PR.

3. **Minor:** the scene-manager branch coverage tests rely heavily on `Object.create(SceneManager.prototype)` + direct prototype-setter invocation and private-field access (`_buildVolume`, `prevStartLayer`, `materials.uniforms`). Achieves the goal, but is white-box and brittle — refactors touching internals will churn these. Acceptable for the coverage lock-in; worth remembering later.

4. Lockfile churn (vite 8.0.16→8.2.2, vitest 4.1.0→4.1.11, lightningcss, rolldown, picomatch) comes along with the install. Benign, just noting.

PR #350, opened 2026-08-21
Precompute the ring of sines and cosines in ExtrusionGeometry

sophiedeziel:
`ExtrusionGeometry.generateSegment` computes the ring of sines and cosines inside the radial loop:

```js
for (let j = 0; j <= radialSegments; j++) {
  const v = (j / radialSegments) * Math.PI * 2;
  const sin = Math.sin(v);
  const cos = -Math.cos(v);
```

That loop runs once per point, on every path. But `j` only ranges over `radialSegments + 1` values and `radialSegments` is fixed for the geometry, so the ring is identical for every point on the path — and for every path in the model.

Hoisting it into a table computed once per geometry.

### Benchmark

3.5 MB 3DBenchy: 163,474 lines → 7023 paths → 3512 extrusion paths → **104,037 points**, generating 537,745 vertices and 2,412,600 indices. `Path.geometry()` passes `radialSegments = 4`.

| | before | after |
| --- | --- | --- |
| trig calls | **1,040,370** | 10 |
| tube geometry generation | 103 ms | **57 ms** |
| speedup | — | **1.81×** |

Measured in Chrome, median of 7 runs after a warm-up, building every extrusion geometry in the model.

An isolated micro-benchmark of just the trig put the saving at ~34 ms; the real end-to-end saving is ~46 ms, so removing the calls helps the surrounding loop as well.

### Correctness

The precomputed values are the same doubles `Math.sin`/`Math.cos` return, stored in a `Float64Array`, so results are bit-identical rather than merely close.

Verified rather than assumed: the previous implementation was reimplemented as a reference and its position and normal buffers compared element-by-element against the new output across **500 geometries — all identical**, no tolerance. The model renders unchanged.

Two regression tests added:

- a four-segment ring around a path along +X lands exactly on the axes and closes back on itself, which pins the table's ordering and sign
- every ring normal stays a unit vector on an arbitrary 3D path

### Scope

This is one of three independent wins found in `ExtrusionGeometry`. The other two — pre-sizing the four `number[]` buffers (~6.7M pushes, worth ~26 ms) and hoisting the five `Vector3` allocated per point (520,185 per render, worth ~4 ms) — are deliberately left out to keep this reviewable. Together all three take geometry generation from 103 ms to 39 ms.

### DIFF
--- a/src/__tests__/extrusion-geometry.ts
+++ b/src/__tests__/extrusion-geometry.ts
@@ -86,3 +86,43 @@ test('ExtrusionGeometry widens its index buffer when the vertex count needs it',
   // and the widest index it wrote is still addressable
   expect(Math.max(...Array.from(wide.index.array.slice(-6)))).toBeLessThan(wide.attributes.position.count);
 });
+
+test('ExtrusionGeometry walks the ring in the order the trig dictates', () => {
+  // Four radial segments land on the axes, so the ring around a path running
+  // along +X is exact: up, left, down, right, back to the start. Any drift in
+  // the precomputed sin/cos table shows up here immediately.
+  const points = [new Vector3(0, 0, 0), new Vector3(1, 0, 0)];
+  const geometry = new ExtrusionGeometry(points, 2, 2, 4);
+
+  const round = (n: number) => {
+    const rounded = Math.round(n * 1e6) / 1e6;
+    return rounded === 0 ? 0 : rounded; // collapse -0
+  };
+  const firstRing = Array.from(geometry.attributes.normal.array.slice(0, 15)).map(round);
+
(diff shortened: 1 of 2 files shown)


remcoder (approved):
Clean, correct micro-optimization. Verified locally (158/158 pass). The precomputed table is behavior-identical: same formula `(j / radialSegments) * Math.PI * 2`, same sign convention (`-Math.cos`), and Float64 matches `Math.sin`/`Math.cos` output precision exactly, so no floating-point drift.

The new tests are well chosen:
- The axis-aligned ring walk is a sharp check — with 4 radial segments and a +X path the expected normals are exact, so any table error fails immediately.
- The unit-vector and identical-rings tests lock in the "same ring for every point" invariant the optimization relies on.

One pre-existing edge (not a regression): `radialSegments = 0` divides by zero in both the old and new code — worth a guard if anyone ever hits it, but out of scope here.

PR #352, opened 2026-08-21
Tokenize G-code lines in a single pass

sophiedeziel:
`parseCommand` tokenizes every line like this:

```js
const parts = cmd
  .split(/([a-zA-Z])/g)   // capturing split: an array with every fragment
  .slice(1)               // a second array
  .map((s) => s.trim());  // a third, plus a new string per token
```

On the 3.5 MB 3DBenchy that runs **163,474 times**, allocating three arrays and a trimmed string per token on each one — before `parseParams` walks the result again in pairs.

The structure it produces is simple: a letter opens a word, and that word's value runs to the next letter. Walking the line once yields the same words without building any of the intermediate arrays.

### Benchmark

3.5 MB 3DBenchy, 163,474 lines. Chrome, median of 5 runs after warm-up. Both sides construct `GCodeCommand` objects, so the comparison is like for like.

| | before | after |
| --- | --- | --- |
| tokenizing all lines | 46.2 ms | **28.2 ms** |
| full `parseGCode` | 75.4 ms | **43.8 ms** |

The full-parse saving (~32 ms) is larger than the tokenizer delta alone (~18 ms), which is consistent with the reduced allocation pressure helping the rest of the parse too.

### Correctness

Every line in the file was parsed both ways and compared on `gcode`, `params`, `comment` and `src`: **0 differences across 163,474 lines**.

That includes the quirks the regex split produced, which are preserved deliberately rather than "fixed":

- `M117 Hello` has no value after each letter, so every letter becomes a `NaN` parameter — `{h: NaN, e: NaN, l: NaN, o: NaN}`
- a line with two semicolons keeps only the span between them as the comment
- an empty comment (`G1 X1 ;`) is `undefined`, not `''`

Changing any of these would be a behaviour change, so they stay, and there are tests pinning them.

15 tests added covering the command, parameters, fractional command numbers (`G28.1`), missing whitespace (`G1X10Y20`), `src` being kept verbatim, blank and comment-only lines, `keepComments: false`, and the three quirks above.

### Also removed

`parseParams` and `isAlpha` are folded into the walk. Both were `private` with no other callers.

### DIFF
--- a/src/__tests__/gcode-parser.ts
+++ b/src/__tests__/gcode-parser.ts
@@ -136,3 +136,77 @@ test('gcode commands without gcode should result in a command with empty string 
   const cmd = parser.parseCommand(gcode) as GCodeCommand;
   expect(cmd.gcode).toEqual('');
 });
+
+describe('parseCommand tokenizing', () => {
+  const parse = (line: string) => new Parser().parseCommand(line) as GCodeCommand;
+
+  test('reads a command and its parameters', () => {
+    const cmd = parse('G1 X100.5 Y-20 E0.04 F1200');
+    expect(cmd.gcode).toEqual('g1');
+    expect(cmd.params).toEqual({ x: 100.5, y: -20, e: 0.04, f: 1200 });
+  });
+
+  test('lower-cases the command and the parameter letters', () => {
+    const cmd = parse('G28 X0');
+    expect(cmd.gcode).toEqual('g28');
+    expect(cmd.params.x).toEqual(0);
+  });
+
+  test('keeps a fractional command number', () => {
+    expect(parse('G28.1 X0').gcode).toEqual('g28.1');
+  });
+
+  test('tolerates missing whitespace between words', () => {
+    const cmd = parse('G1X10Y20');
+    expect(cmd.gcode).toEqual('g1');
+    expect(cmd.params).toEqual({ x: 10, y: 20 });
+  });
+
+  test('keeps the src line verbatim', () => {
+    const line = '  G1 X1 ; move  ';
+    expect(parse(line).src).toEqual(line);
+  });
+
+  describe('comments', () => {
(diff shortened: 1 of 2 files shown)


remcoder (approved):
Solid refactor, behavior-preserving. Verified two ways:

1. Ran the PR's tests locally — 167/167 pass.
2. Wrote a differential test running the OLD parser (regex-split + parseParams, checked out from the base commit) against the NEW single-pass tokenizer over 25 edge-case lines — blank lines, comment-only lines, double/triple semicolons, M117-style free text, missing whitespace, fractional commands, trailing values, `keepComments=false`. Both parsers produced byte-identical GCodeCommand objects on every line. The "treats letters in free text as parameters, as it always has" claim is accurate — e.g. `M117 Hello world` yields `{h,e,l,o,w,r,d}` all NaN in both.

The single-pass walk is a faithful translation of the old split logic: the letter-opens-word scan matches `split(/([a-zA-Z])/g)` semantics, and `indexOf(';')` slice matches `split(';')[1]` (including the second-semicolon truncation). The `isAlphaCode` hoist to a module-level function is clean and now pure.

Notes (non-blocking):
1. `isFirstWord` gates the gcode, so a line like `X100 Y200` (no leading G/M) parses `x` as gcode — same as before (`parts[0]` was the gcode). Covered by existing behavior; just worth knowing.
2. `cmd` is now a `slice()` instead of a `split()[0]`, so a line with a semicolon no longer builds an array — that's the point, but it does mean the comment extraction runs `indexOf` twice on long lines. Negligible.
3. The `Number(value)` in the gcode branch vs `parseFloat(value)` for params: `Number('28.1')` is 28.1, `Number('')` is 0 — matches the old `Number(parts[1])` exactly (the old code also used `Number` for the command and `parseFloat` for params). Good parity.

PR #353, opened 2026-08-21
Make source line retention opt-in, and count lines separately

sophiedeziel:
`Parser` keeps every source line on `lines`. For the 3.5 MB 3DBenchy that is 163,474 strings — roughly **5 MB**: 3.37 MB of characters, a 1.25 MB pointer array, and a header per string. Nothing in the library reads the text back.

Retention is now opt-in:

```js
new Parser({ keepLines: true });
new GCodePreview({ canvas, keepLines: true });   // reaches the same option
```

Off by default. The idea is that the lines could be useful, for clients of the library or for internal features. So I'm not a fan of dropping them unconditionally. I don't want to create a one-way door. The default could dynamically decide if the lines are needed or not if a set of features enabled requires them.

### Two bugs fixed along the way

**`lines` only ever held the last chunk.** `parseGCode` did `this.lines = input.split('\n')` — an assignment, not an append — and a streaming parse calls it once per chunk. After the demo's streamed load the parser held **72,055 of 163,474 lines**. So the field could not have served the "keep the source" purpose it existed for, on exactly the large files where it matters. It now appends.

`parseGCode` also now builds commands from the current chunk rather than the accumulated buffer, so a streaming parse does not re-tokenize what it has already handled.

**The dev GUI's line counter reported that same truncated number**, because it read `parser.lines.length`. It now reads a new `lineCount`, which accumulates across calls and is tracked whether or not the lines are kept. On the Benchy it reports **163,474** where it previously showed **72,055**.

`lineCount` and `keepLines` are independent — you get an accurate count without paying to keep the text.

### Verified

| | `lines` held | `lineCount` |
| --- | --- | --- |
| default, streamed load | **0** | 163,474 ✅ |
| `keepLines: true`, whole file | 163,474 | 163,474 |
| `keepLines: true`, parsed in 2 chunks | **163,474** | 163,474 |

The chunked row is the fix — previously it would have held only the second chunk.

Dev GUI checked in the browser: the `lines.count` widget renders `163474`. The old binding would now show `0`.

Heap deltas for the retained lines measured at 5.8 / 1.7 / 5.1 MB across three runs — too noisy to quote precisely without forced GC, hence "roughly 5 MB" from the deterministic parts rather than a false-precision figure.

### Breaking

`parser.lines` is empty unless `keepLines` is set. It is a public field in the published typings, so anyone reading it needs the option — though on a streamed load they were getting a partial file anyway.

The `preserving-parser` tests, whose whole subject is this capability, now opt in, and there are new tests for the default being empty, for accumulation across chunks, and for `lineCount`.

### Not addressed here

Peak memory during a non-streaming parse is ~30 MB for a 3.5 MB file, because `parseGCode` materialises all 163,474 `GCodeCommand` objects at once. That dwarfs the retention fixed here, but changing it means changing `parseGCode`'s shape, which deserves its own discussion.

`GCodeCommand.src` is public, has no readers anywhere in the repo, and costs nothing to retain — it references the same string as `lines[i]`, on an object that is discarded right after interpretation. Left alone deliberately.

### Review follow-up

- Fixed the streaming off-by-one: `readStream`'s carried tail started at the newline itself, so every chunk boundary fed a spurious empty line into `parseGCode` and inflated `lineCount`. The fix is one character (`slice(idxNewLine + 1)`), but it changes behavior beyond `keepLines`: the interpreter no longer receives one empty command per chunk boundary.
- Extracted the per-chunk split into `splitChunk` (`src/helpers/split-chunk.ts`), a pure helper `readStream` now uses, so boundary behavior is testable without constructing a `GCodePreview` (WebGLRenderer throws under happy-dom). Its known no-newline-chunk quirk is preserved unchanged.
- Regression test drives `splitChunk` + `Parser({ keepLines: true })` across a mid-line chunk boundary: no empty line appears and `lineCount` matches.
- Trailing-newline semantics (`split('\n')` yields one final empty line) are now documented on `lines`/`lineCount` rather than changed.

### DIFF
--- a/src/__tests__/preserving-parser.ts
+++ b/src/__tests__/preserving-parser.ts
@@ -1,30 +1,94 @@
-import { test, expect } from 'vitest';
+import { test, expect, describe } from 'vitest';
 
 import { Parser } from '../gcode-parser';
 
-test('all input should be preserved', () => {
-  const parser = new Parser();
-  const gcode = `G1 X0 Y0 Z1 E1`;
-  const parsed = parser.parseGCode(gcode);
-  expect(parsed).not.toBeNull();
-  const unparsed = parser.lines.join('\n');
-  expect(unparsed).toEqual(gcode);
+describe('with keepLines', () => {
+  test('all input should be preserved', () => {
+    const parser = new Parser({ keepLines: true });
+    const gcode = `G1 X0 Y0 Z1 E1`;
+    const parsed = parser.parseGCode(gcode);
+    expect(parsed).not.toBeNull();
+    const unparsed = parser.lines.join('\n');
+    expect(unparsed).toEqual(gcode);
+  });
+
+  test('multiple lines should be preserved', () => {
+    const parser = new Parser({ keepLines: true });
+    const gcode = `G1 X0 Y0 Z1 E1\nG1 X10 Y10 E10`;
+    const parsed = parser.parseGCode(gcode);
+    expect(parsed).not.toBeNull();
+    const unparsed = parser.lines.join('\n');
+    expect(unparsed).toEqual(gcode);
+  });
+
+  test('comments should be preserved', () => {
+    const parser = new Parser({ keepLines: true });
+    const gcode = `G1 X0 Y0 Z1 E1; this is a comment`;
+    const parsed = parser.parseGCode(gcode);
+    expect(parsed).not.toBeNull();
+    const unparsed = parser.lines.join('\n');
+    expect(unparsed).toEqual(gcode);
+  });
+
+  test('input parsed in chunks should be preserved whole', () => {
+    // A streaming parse calls parseGCode once per chunk. Preserving only the
+    // most recent chunk would make the option useless for the large files it
+    // exists for.
+    const parser = new Parser({ keepLines: true });
+    parser.parseGCode('G1 X0 Y0\nG1 X1 Y1');
+    parser.parseGCode('G1 X2 Y2\nG1 X3 Y3');
+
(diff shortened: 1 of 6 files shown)


remcoder (approved):
Clean, well-scoped behavioral change. Verified locally: 162/162 tests pass, typeCheck and lint both clean.

Checked the blast radius of making `lines` opt-in:
- Only consumer of `parser.lines` in the repo was `dev-gui.ts`, which you updated to `lineCount`. The demo reads only `parser.metadata.thumbnails`. No README/docs references. Nothing breaks.
- The thumbnail merge loop is preserved byte-for-byte from the old `parseGCode`, so metadata accumulation across streaming chunks still works.
- The streaming semantics are an improvement over the old code, which *overwrote* `this.lines` every call (and held the caller's array by reference for array input). Now with `keepLines` it appends copies, and the chunk-is-parsed-once test pins the "don't redo handled chunks" invariant.

Notes (non-blocking):
1. This is technically a breaking API change for any external consumer reading `parser.lines` — fine for an alpha (3.0.0-alpha.4), but the CHANGELOG/release notes should mention it.
2. The `parseGCode` JSDoc still says it "maintains the original source lines", which is now only true with `keepLines`. Worth a one-line tweak to avoid confusion.
3. The `lineCount` accumulation across calls is a deliberate divergence from the old reset-per-call behavior — the tests document it well, so this is fine, just noting it's a semantic change people might trip on.
4. Good call on the `push` loop instead of spread — `push(...lines)` would throw on very large chunks.

Approve.

PR #362, opened 2026-08-22
Share the precomputed sin/cos ring across ExtrusionGeometry instances

### LINKED ISSUE #360: Share the precomputed sin/cos ring across ExtrusionGeometry instances
Hoist the per-instance sin/cos ring table in `extrusion-geometry.ts` to a module-level cache keyed by `radialSegments` (always 4 in practice), so all instances share one precomputed ring.

Suggested in [this review comment](https://github.com/xyz-tools/gcode-preview/pull/350#issuecomment-5376343166) on #350 — includes a code sketch.

- Deliberately deferred until #349/#350/#351 (which all touch `extrusion-geometry.ts`) are merged, to avoid more conflicts.
- Add a cross-instance test (two geometries, same `radialSegments`, identical rings) and a comment noting the cached arrays are shared and must never be mutated.
- Expected impact is code cleanliness, not measurable perf.

sophiedeziel:
Hoists the per-instance `ringSin`/`ringCos` tables in `extrusion-geometry.ts` to a module-level cache keyed on `radialSegments`, per [remcoder's suggestion on #350](https://github.com/xyz-tools/gcode-preview/pull/350#issuecomment-5376343166). The constructor now pulls the shared tables from `ringTrig(radialSegments)`; `generateSegment` is untouched.

**Why:** every `ExtrusionGeometry` rebuilt an identical 5-element trig table (all call sites pass `radialSegments = 4`), so a large model allocated and refilled the same two `Float64Array`s hundreds of thousands of times. Now the table is built once per distinct `radialSegments` value. This is a code-cleanliness change — the expected perf impact is negligible, per the prior review discussion.

**Safety:** the cached arrays are shared across instances and are read-only by contract; a comment on the cache spells that out. A new cross-instance test pins the sharing (two geometries, same `radialSegments`, identical first rings), and a different-`radialSegments` test covers the cache-miss branch, keeping `extrusion-geometry.ts` at 100% on all coverage axes.

Closes #360


remcoder (approved):
LGTM

PR #407, opened 2026-08-24
Finish the port of progressive rendering

sophiedeziel:
## What

The model now draws **live while a G-code stream is being read** — the print grows on screen as chunks arrive, instead of the canvas staying empty until the stream ends.

| ~20% streamed | ~80% streamed | complete |
|---|---|---|
| early | mid | complete |

*(3DBenchy preset streamed over emulated Fast 3G with a cold cache; screenshots taken mid-stream in Chrome, with `highlightTopLayer`/`highlightLastSegment` on and contrasting colors — orange top layer, blue last segment — against the model's mint-green. The orange ring visibly climbs the hull as the print streams in, well before the model is finished — layer 78/240 and 113/240 above, vs. the finished chimney at 240/240.)*

## How it works

As chunks arrive from the network, `readStream` parses and executes them into the job as before — the new part is that between chunks it now asks the scene to draw what has accumulated so far, instead of waiting for the stream to end:

```mermaid
sequenceDiagram
    participant N as Network stream
    participant R as GCodePreview.readStream
    participant J as Job
    participant S as SceneManager
    participant O as ObjectsManager

    loop each chunk
        N->>R: decoded text chunk
        R->>J: parse + execute (paths accumulate)
        opt ≥ liveRenderInterval since last draw
            R->>S: renderProgressive()
            S->>O: renderPaths(all paths except the last)
            Note over O: builds only the not-yet-drawn tail<br/>as one new geometry batch
        end
    end
    R->>S: renderAnimated()
    Note over S,O: draws whatever remains,<br/>including the held-back final path
```

The reason this is cheap is a contract the `ObjectsManager` already had: it remembers every path it has built (`renderedPaths`), so progressive rendering can hand it a *growing prefix of the same list* on every call and each path is built exactly once — each `renderProgressive()` call materializes only the paths that arrived since the previous one, as a single geometry batch. Nothing needs to present the result explicitly either: the continuous `animate()` loop is already rendering the scene every frame, so new geometry simply appears on the next frame after it is added.

Two decisions shape what each progressive draw is allowed to touch:

```mermaid
flowchart LR
    A[paths in the job] --> B{already built?}
    B -- yes --> C[skipped<br/>renderedPaths]
    B -- no --> D{is it the job's last path?}
    D -- yes --> E[held back<br/>it may still be growing]
    D -- no --> F[built into this draw's batch]
```

- **Throttling.** A path is built per chunk arrival, but chunks can arrive far faster than 60 fps, and every draw adds a geometry batch that costs a draw call for the rest of the session. So draws are rate-limited to one per `liveRenderInterval` milliseconds (default **250** — the throttled ~22 s test stream produced 83 draw calls). Lower values grow the model more smoothly; `0` draws on every chunk, effectively once per frame.
- **The held-back tail.** The interpreter keeps appending vertices to the job's last path across chunk boundaries, but a path is only ever *built* once — drawing it early would freeze it with whatever vertices it happened to have at that moment, and its later vertices would never appear. Browser verification caught exactly this: without the holdback the final scene was missing precisely the tail geometry (789,608 vs 803,608 triangles). Each draw therefore stops one path short of the end; the final `renderAnimated()` at stream end picks up whatever remains, so the finished scene is identical to a non-progressive render.

`renderProgressive()` is tagged `@internal` (like `animate()`) — it only exists so `readStream` can call it across the class boundary; consumers keep using `processGCodeStream`, whose `render: false` option now flows down into `readStream` and skips live drawing entirely.

## Fix: the top-layer/last-segment highlight now tracks the stream live

`highlightTopLayer`/`highlightLastSegment` originally only redrew on a full render, so with a stream still coming in the highlight stayed off until the whole model finished — "top layer" and "last segment" meant the ones a *finished* print settles on, not the ones most recently *received*, which is what a live view should show.

`SceneManager` now re-evaluates the highlight on every `renderPaths()` call, including progressive ones, remembering which layer and path it last drew on so it can revert the highlight it's moving off of before drawing the new one. The split follows the policy/mechanism boundary between the two classes: `SceneManager` keeps only that memo (deciding *when* the top layer moved needs job and end-layer knowledge), while `ObjectsManager` — which already owns the groups, disposables and rendered-path bookkeeping — tracks what the highlight actually drew and claimed, and undoes exactly that in its new `revertHighlight()` (unclaim + remove + dispose). Tracking per actual draw/claim matters: reverting whole-layer path lists would also unclaim paths the per-tool batches drew normally (the `lastSegmentColor`-without-`topLayerColor` case), and the next progressive draw would then draw them a second time, stacking duplicate geometry each time the highlight moved — a regression test pins this. Splitting a path into its final segment and body moved onto `Path.splitLastSegment()`, since vertex slicing is path-domain work with no dependence on camera, scene or policy.

A structurally new top layer reverts the old highlight even before the new layer has anything visible yet (its only path may be the streaming-safe window's held-back tail) — otherwise the previous layer would stay highlighted for as long as the new one takes to produce something to draw. "Top layer" selection still keys off `job.layers` rather than raw path indices, so a structurally-empty trailing layer or a non-planar job correctly shows no highlight instead of falling back to an earlier one — matching the pre-existing full-render behavior. Layer identity alone isn't enough, though: the layer's own paths are intersected with the paths visible to *this specific render call*, so `renderAnimated`'s frame-by-frame reveal of an already-complete job doesn't jump the highlight ahead of the per-tool draw it's supposed to follow — caught by a dedicated test before it shipped.

## Browser evidence

Fast 3G, cold cache, 3.7 MB 3DBenchy (measured through `window._preview`):

| Time | Paths in job | On-screen triangles |
|---:|---:|---:|
| mid-stream (early) | 1,378 | 180,008 |
| mid-stream (later) | 5,696 | 731,472 |
| complete | 7,023 | **803,608** — identical to the non-progressive baseline |

Before this change the on-screen triangle count stayed **0** for the entire stream.

The screenshots live on the orphan branch `pr-assets-live-stream-rendering`, safe to delete after merge.

### DIFF
--- a/src/scene-manager.ts
+++ b/src/scene-manager.ts
@@ -11,7 +11,6 @@ import { Path, PathType } from './path';
 import {
   Color,
   ColorRepresentation,
-  Object3D,
   OrthographicCamera,
   PerspectiveCamera,
   REVISION,
@@ -131,10 +130,6 @@ export class SceneManager {
   private _highlightedLayerIndex?: number;
   /** The path the highlight last treated as the layer's last-received path */
   private _highlightedLastPath?: Path;
-  /** Scene objects the current highlight drew, so they can be reverted */
-  private _highlightObjects: Object3D[] = [];
-  /** Paths the current highlight claimed, so they can be unclaimed on revert */
-  private _highlightedPaths: Path[] = [];
   /** Last render time in milliseconds */
   lastRenderTime = 0;
   /** Whether to render in wireframe mode */
@@ -908,9 +903,7 @@ export class SceneManager {
     // (including both being empty, e.g. a non-planar job with no layers at all)
     if (topLayerIndex === this._highlightedLayerIndex && lastPath === this._highlightedLastPath) return;
 
-    this.objectsManager.revertHighlight(this._highlightObjects, this._highlightedPaths);
-    this._highlightObjects = [];
-    this._highlightedPaths = [...layerExtrusions];
+    this.objectsManager.revertHighlight();
     this._highlightedLayerIndex = topLayerIndex;
     this._highlightedLastPath = lastPath;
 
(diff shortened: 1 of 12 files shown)


src/scene-manager.ts
```diff
@@ -126,6 +127,14 @@ export class SceneManager {
   private _topLayerColor?: Color;
   /** Last segment color */
   private _lastSegmentColor?: Color;
+  /** Layer index the top-layer/last-segment highlight is currently drawn on */
+  private _highlightedLayerIndex?: number;
+  /** The path the highlight last treated as the layer's last-received path */
+  private _highlightedLastPath?: Path;
+  /** Scene objects the current highlight drew, so they can be reverted */
+  private _highlightObjects: Object3D[] = [];
```
sophiedeziel: Note to self: this belongs in the objects-manager

PR #443, opened 2026-09-02
Run workflow on merge

remcoder:


.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -5,6 +5,7 @@ name: Deploy to Firebase Hosting on PR
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened, closed]
```
sophiedeziel, line 8:
```suggestion
    types: [opened, synchronize, reopened]
```

Could skip the whole workflow when closing. Teardown is not supported by the official action, but found this where people shared their solution: https://github.com/FirebaseExtended/action-hosting-deploy/issues/60

sophiedeziel: If the expiration is for a preview specifically (not global), an alternative would be to have 30 days for all, and close sets it to 1 day so it cleans itself faster.

.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -23,7 +24,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - run: npm ci && npm run predeploy && npm run typedoc
+      - if: github.event.action != 'closed'
+        run: npm ci && npm run predeploy && npm run typedoc
```
sophiedeziel, line 28: Asked copilot how to only run when changes target actual code. Did not try it, but sounds legit

```suggestion
      - id: filter
              uses: dorny/paths-filter@v3
              with:
                filters: |
                  target:
                    - 'src/**'
                    - '!src/__tests__/**'
                    - 'demo/**'
      - if: github.event.action != 'closed'
        run: npm ci && npm run predeploy && npm run typedoc
```

.github/workflows/firebase-hosting-pull-request.yml
remcoder: oops, opus was hallucinating then.. :-P

Issue #402, opened 2026-08-24
3.0.0 release checklist


sophiedeziel:
Pre-release checklist for **3.0.0 stable**

## Release gate

- [ ] **Benchmarks 2.x vs 3.0** — same machine/browser/files for both versions; measure parse time, time-to-first-render, FPS while orbiting, and peak memory. Post results as a comment here.
- [ ] **Large prints still work** — bundled demo gcodes plus one genuinely large file (≥50 MB), both drag-and-drop and streaming, tubes and lines.
- [ ] **Slow computer test** — DevTools CPU throttling (4x/6x) and ideally one real low-end machine; stays usable, no crashes or context loss.
- [ ] **Framework wrappers** — React, Vue, Svelte examples build and run against 3.0; mount, prop update, and unmount (`dispose()`) all clean, incl. React StrictMode double-mount.
- [ ] **README up to date** — code samples, options list, and links all match the 3.0 API.
- [ ] **Memory leak test** — ~20 create/load/dispose cycles: heap and `renderer.info.memory` return to baseline, no WebGL context accumulation, listeners removed.

## Also worth doing

- [x] Release notes with breaking changes + migration guide link
- [ ] Version bump + `npm publish --dry-run` sanity check per `PUBLISHING.md`
- [ ] Verify demo deploy + regenerate API docs
- [ ] Sweep issues/PRs labeled `3.0` / `port to v3`

remcoder:
see #442 for larger files

sophiedeziel:
> Slow computer test — DevTools CPU throttling (4x/6x) and ideally one real low-end machine; stays usable, no crashes or context loss.

I just tested on my 2019 macbook pro with 20x CPU throttling. It is quite slow to render the benchy but not unreasonable even for 20x. Though, this laptop has a reasonable GPU, which also comes into play, and 32GB of RAM. Not the definition of low end at all, just old. 

I don't have access to a low-end machine right now, but it's worth considering that I think 3.0 performs better than 2.8 on most metrics (memory is most likely worse though).

I'll benchmark the comparison soon.

sophiedeziel:
Used the benchmark tool prototyped in #453 

Version A: 2.8.0
Version B: What's on the `develop` branch at the time of this writing, last commit is https://github.com/xyz-tools/gcode-preview/commit/2b4e68c1d5a5cd9ab9ae85fbb8bf9f5e2d5d2401

## Benchy, tube rendering

[...]

PR #454, opened 2026-09-06
Sync docs with the current state of develop

sophiedeziel:
Ref #402

An audit of README.md, CONTRIBUTING.md and PUBLISHING.md against the current code found the docs had drifted. This PR brings them back in line, following a few principles:

- **Docs should match what the code actually does** — every claim (commands, options, CI checks, release steps) was verified against the source and workflows, and corrected where stale.
- **Document the traps** — anything where following the old docs would fail (coverage gate, test file naming, PR base branch) is now called out explicitly.
- **Write for the 3.0 release** — the README describes the current API as *the* API, with no prerelease framing.
- **Don't duplicate** — the README defers to CONTRIBUTING.md and the API docs instead of restating them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,12 @@ newest supported version; after merge, a matrix job re-runs the suite against
 every supported three.js release. Avoid APIs that aren't available across that
 whole range.
 
+## AI usage
+
+AI-assisted contributions are welcome, but usage of AI tools (Copilot, Claude,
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -14,6 +14,8 @@ Other things that are always helpful:
 
 ## Development setup
 
+Use Node 22 (`.nvmrc` pins `v22.12.0`, so `nvm use` picks it up). CI builds and tests on Node 22.
+
```
sophiedeziel:
```suggestion
```

PR #456, opened 2026-09-07
feat: support G92.1 coordinate offset resets

### LINKED ISSUE #424: Support G92.1 (reset coordinate system offsets)
Clear offsets set by G92 (appears in Bambu and Smoothieware output).

**Seen in:** Bambu machine start gcode; Smoothieware.
**Behavior:** reset the logical-position offsets established by G92 without moving.
**Related:** #433

Docs: [RepRap wiki G-code](https://reprap.org/wiki/G-code), [Smoothieware supported G-codes](https://smoothieware.org/supported-g-codes)

kudala-bharani:
## Command

Support `G92.1`. Closes #424.

## Dialect & reference

This follows [Marlin's G92.1 implementation](https://github.com/MarlinFirmware/Marlin/blob/922aa6619128928da3523f88c528d71c49756169/Marlin/src/gcode/geometry/G92.cpp#L36-L66): clear XYZ workspace offsets and leave E unchanged. E is set directly by G92, not stored in a workspace offset.

The previous revision introduced a second, physical-E model. That was unnecessary and conflicted with the logical E tracking being added in #461, so it has been removed. `G1 E10 / G92 E0 / G92.1` now leaves E at zero; a nonzero G92 E value is preserved too. Current [Smoothieware source](https://github.com/Smoothieware/Smoothieware/blob/38e2cc083db0e4f768535a9bf2d32cdf104ea980/src/modules/robot/Robot.cpp#L624-L674) also limits its E reset to ordinary G92 (`subcode == 0`).

## Behavior and state

G92.1 was previously ignored. It now zeros `state.positionShift.x/y/z`, so subsequent linear and arc targets use unshifted coordinates. It does not move the printhead, change E, home unknown axes, or add geometry/statistics. Existing G92 and move handling are unchanged from develop.

M82/M83 support remains in #461; that unfinished branch is not included here. G92.2/G92.3 remain unhandled. The new `resetPositionShift` command-barrel export follows the existing handler-registration pattern; no State fields are added.

## Checks

On Node 22.23.2: **872 tests pass, with 100% coverage**. Type checking, lint, and build pass (the existing `lil-gui` global-name warning remains).

Five reset cases failed against the previous E-offset revision and pass now. All 18 reset cases run through the parser and interpreter, including chunked execution and E0/nonzero E preservation. Extrusion cases are consolidated in the reset suite.

An isolated combination with #461 also passed 24 targeted tests, including absolute/relative E checks. That check required combining the additive registry, import, and README edits; no mode-handling code needed changing.

Assisted by OpenAI Codex.


src/interpreter.ts
```diff
@@ -72,8 +74,9 @@ export const handlers: ReadonlyMap<string, CommandHandler> = new Map<string, Com
  * This class looks up each command in the handler registry and executes it,
  * translating commands into movements and state changes in the print job. It
  * supports common G-code commands including linear moves (G0/G1), arcs (G2/G3),
- * unit changes (G20/G21), homing (G28), position setting (G92), and tool
- * selection. Commands without a registered handler are ignored.
+ * unit changes (G20/G21), homing (G28), position setting (G92), workspace
+ * offset reset (G92.1), and tool selection. Commands without a registered
+ * handler are ignored.
```
sophiedeziel: At this point, I think we can stop enumerating every single command supported in the comment 😄 The code is already very self-explanatory

src/interpreter/commands/reset-position-shift.ts
```diff
@@ -0,0 +1,9 @@
+import type { CommandHandler } from '../../interpreter';
+
+/** Clears G92 workspace offsets without moving the printhead or resetting E. */
+export const resetPositionShift: CommandHandler = (command, job) => {
+  const { positionShift } = job.state;
+  positionShift.x = 0;
+  positionShift.y = 0;
+  positionShift.z = 0;
```
sophiedeziel, line 12: The documentation states that this command should also reset the extrusion offset : https://smoothieware.org/supported-g-codes

However, I realize that a `positionShift` was not implemented for `e` in `setPosition`, it's only resetting it to 0. 

If you are willing to add that as part or this PR, that would be super helpful, as this command would do exactly what it is supposed to do.

src/interpreter.ts
kudala-bharani: Addressed in a15e36b: removed the supported-command enumeration and kept the short explanation of registry dispatch.

src/interpreter/commands/reset-position-shift.ts
kudala-bharani: Added in a15e36b. `state.e` now tracks physical extrusion through linear, arc, and E-only moves; `G92 E` stores the logical offset in `positionShift.e`, and `G92.1` clears it without moving the extruder. Bare and partial G92 commands are covered too.

The existing relative-E interpretation, geometry, and statistics are preserved; M82/M83 mode switching remains outside this PR. The README and PR description now explain the physical/logical E distinction.

All 14 additional regression cases failed before this follow-up and pass afterward. The full suite passes 870 tests with 100% coverage; type checks, lint, and build pass on Node 22. Ready for another look.

src/__tests__/interpreter/commands/extrusion-position.ts
sophiedeziel, line 1: This should not be a new file, the new tests should complement the existing test files targeting `set-position.ts` or `reset-position-shift.ts`.

README.md
```diff
@@ -109,10 +109,15 @@ The interpreter currently handles:
...
 | `G92` | set position |
+| `G92.1` | reset coordinate system offsets |
 | `T0`–`T7` | tool selection |
 
 Commands without a handler are parsed but ignored by the interpreter.
 
+`G92` stores offsets for X, Y, Z, and extrusion (E); `G92.1` clears them without moving.
+`state.e` tracks cumulative physical extrusion, with logical position `state.e - state.positionShift.e`.
+Extrusion moves still use relative E values; `M82`/`M83` mode switching is not yet implemented.
+
```
sophiedeziel: This is too much details for the general Readme. Documentation for gcode exists is many sources, we should not maintain a copy of that documentation here

src/interpreter/commands/linear-move.ts
```diff
@@ -12,6 +12,9 @@ import type { CommandHandler } from '../../interpreter';
  */
 export const linearMove: CommandHandler = (command, job) => {
   const { x, y, z, e, f } = command.params;
+  const { state } = job;
+  // E is currently relative, so coordinate offsets do not affect its delta.
+  state.e += e ?? 0;
```
sophiedeziel: state.e += e ?? 0 (also arc-move.ts:67) has no M82/M83 tracking, but state.ts documents e as “physical extrusion position”. Absolute E coordinates get summed as if they were deltas.

Example: Cura default M82 / G1 X10 E5 / G1 X20 E10 / G92 E0 / G1 X30 E12 → printer extruded 22, Marlin’s E is 12, but state.e = 27. The documented logical E = state.e − positionShift.e is wrong for most real slicer output.

Can you add to the comment that M82 and M83 are not implemented yet

src/interpreter/commands/reset-position-shift.ts
```diff
@@ -0,0 +1,10 @@
+import type { CommandHandler } from '../../interpreter';
+
+/** Clears all G92 offsets without moving the printhead or extruder. */
+export const resetPositionShift: CommandHandler = (command, job) => {
+  const { positionShift } = job.state;
+  positionShift.x = 0;
+  positionShift.y = 0;
+  positionShift.z = 0;
+  positionShift.e = 0;
```
sophiedeziel: Problem: zeroing positionShift.e while keeping physical state.e makes logical E jump by the old offset — matching no firmware, and none is cited.

Example: G1 E10 / G92 E0 / G92.1 → Marlin’s G92.1 never touches E, Smoothieware zeroes extruder position — both leave logical E at 0; this leaves it at 10.

Recommendation: cite one firmware in the doc comment; if Smoothieware is the model, use positionShift.e = state.e (logical E → 0). Note G92.2/G92.3 unhandled in the README.

src/interpreter/commands/linear-move.ts
sophiedeziel: Update: I'm adding M82 and M83 here: https://github.com/xyz-tools/gcode-preview/pull/461

This will enable a much better resolution of this

kudala-bharani: Thanks for taking that on. I've dropped my E accumulator and kept the existing `state.e` model so this doesn't duplicate #461. I tried the two branches together locally, including M82 and M83; the tests pass after combining the imports, registry entries, and README additions. Your unfinished branch isn't included in this PR.

src/__tests__/interpreter/commands/extrusion-position.ts
kudala-bharani: Moved the extrusion cases into `reset-position-shift.ts` and removed the separate test file.

README.md
kudala-bharani: Trimmed this back to the command table and a short note that G92.2/G92.3 aren't supported.

src/interpreter/commands/reset-position-shift.ts
kudala-bharani: There's a wrinkle here: [current Smoothieware](https://github.com/Smoothieware/Smoothieware/blob/38e2cc083db0e4f768535a9bf2d32cdf104ea980/src/modules/robot/Robot.cpp#L624-L674) only resets E for plain G92, not G92.1. I've gone with [Marlin's behavior](https://github.com/MarlinFirmware/Marlin/blob/922aa6619128928da3523f88c528d71c49756169/Marlin/src/gcode/geometry/G92.cpp#L36-L66): clear XYZ and leave E alone. The extra E-offset model is gone, so your example now keeps E at 0; nonzero E values are preserved too.

PR #472, opened 2026-09-07
Keep layer metadata whole when a chunk splits its comment block

### LINKED ISSUE #445: [P1] Streaming loses layer metadata when a chunk splits a layer comment block
Split Prusa input between ;LAYER_CHANGE and its ;Z:0.2 / ;HEIGHT:0.2. Expected: both stream layouts retain layers at Z0.2 and Z0.4. Actual: the split loses the first Z/height and collapses both layers into one. The whole-file control passes; the split-block regression fails.

## Reproduction

Deliver a PrusaSlicer header and `;LAYER_CHANGE` as the first stream chunk. Deliver `;Z:0.2`, `;HEIGHT:0.2`, first-layer moves, and a complete second layer at Z0.4 in the next chunk. Compare with the identical text delivered in one chunk. The test fixture defines the exact input and checks metadata and indexed paths.

## Verified result

Expected: both ingestion modes retain layers at Z0.2 and Z0.4 with height 0.2. Observed: the split block loses the first Z/height and both extrusion paths land in one layer. Whole-file control passes; split-block regression fails.

## Regression test

Branch: `codex/p1-stream-layer-metadata-repro` (based on `c99c9db`).

[Reproduction test](https://github.com/xyz-tools/gcode-preview/blob/codex/p1-stream-layer-metadata-repro/src/__tests__/p1-reproduction.ts)

After checking out the branch and installing the project dependencies, run:

```sh
NODE_OPTIONS=--no-experimental-webstorage npm test -- src/__tests__/p1-reproduction.ts
```

The environment flag avoids Node 26’s experimental web-storage conflict with happy-dom. No browser is needed. Only the GPU renderer is stubbed where necessary; parsing, interpretation, and job state remain real. The branch contains a failing regression test and no implementation fix.

sophiedeziel:
Fixes #445.

## The bug

A slicer describes a layer in a run of comments — `;LAYER_CHANGE`, then `;Z:`, then `;HEIGHT:` — and a streamed chunk ends wherever the bytes ran out. When a boundary fell inside that run, the parser read the first half, reported a layer with no Z and no height, and never saw the values that arrived in the next chunk. `LayersMetadataIndexer` then filed every later path into it.

## The fix

`splitChunk` already exists to stop a boundary landing inside a *line*. A run of comments is never the end of anything either, so it now carries one into the next chunk the same way, and the G-code line that closes the run brings it along. `readStream` already flushes whatever the tail holds when the stream ends, so a file finishing on comments needs nothing new.

**Nothing that reads the comments changes.** All four slicer parsers, the metadata contract and `Parser` are untouched — they keep seeing whole blocks, exactly as they do when the whole file is parsed at once.

One detail worth a look: the held run keeps the newline that terminated it. Only the partial line ends a tail without one, because it is the line the next chunk continues; dropping the terminator would glue the next chunk's first characters onto the last held line.

## What it covers

Sweeping **every byte boundary** of the five slicer fixtures and comparing against a single-shot parse:

| dialect | before | after |
|---|---|---|
| PrusaSlicer | 38 | **0** |
| OrcaSlicer | 38 | **0** |
| Slic3r | 32 | **0** |
| Simplify3D | 0 | 0 |
| Cura | 41 | 18 |

149 mismatches down to 18, and the remaining ones are all Cura. Cura is the one dialect this cannot help: it reads a layer's Z from the *move* that follows the marker rather than from a comment, so a boundary between the two is not one `splitChunk` can close. That is a real bug, but a different one from #445 — I will file it separately rather than grow this PR to cover it.

On the five bundled demo files, streamed at 256 B through 64 KB, the layer metadata is now identical to a single-shot parse at every size, with no Z-less layers.

## Tests

- **`split-chunk.ts`** covers the holdback directly, including the newline detail and 1-byte chunking.
- **`slicer-fixtures.ts`** sweeps every byte boundary of four fixtures through `splitChunk`, as a stream really goes, and compares against the single-shot parse. It fails on three of the four without this change.
- **`ingestion-equivalence.ts`** gains a PrusaSlicer fixture. Every fixture in that harness was bare gcode, so `detectSlicer` returned `null` and the layer metadata path never ran in any streaming mode — which is how this survived a harness built to catch exactly this.

949 tests pass, coverage stays at 100% with no new pins, typecheck and lint clean.

## Before / after

`demo/gcodes/calicat.gcode` (PrusaSlicer, 620 KB) streamed in 1 KB chunks — what a slow connection or a serial feed delivers. Same file, same camera, view limited to the bottom third of the detected layers.

**Before** — a boundary lands inside a `;LAYER_CHANGE` block, that layer loses its Z, and everything above collapses into it: 29 layers instead of 174, 2900 paths in one.

before

**After** — 174 layers, 122 paths in the busiest one, identical to parsing the same bytes in one shot.

after

At the demo's default ~64 KB chunks none of the bundled files reproduce this, which is why it went unnoticed: a boundary has to land inside a three-line block.


sophiedeziel:
Reworked the approach after review discussion.

The earlier version kept the parsers as pure functions and had `Parser` **replay** the previous chunk's tail, dropping and reinstating whatever the replayed span had already reported. It worked and was well covered, but it had a flaw worth avoiding: replay-safety became an invisible, per-accumulator obligation. `parseSlicerMetadata` returns both `layers` and `extrusionDimensions`, and each had to implement drop-and-reinstate independently with nothing enforcing it — a trap that had already fired once during the rebase onto #458.

Deferring instead of replaying removes the obligation rather than documenting it: nothing is reported until it is complete, so nothing is ever double-counted and no consumer can read a value that is about to be revised. It also turns out to be strictly stronger. Replay was only correct within its window, so at 256-byte chunks calicat still produced six Z-less layers; deferring has no window and produces none at any chunk size from 256 B to 64 KB.

The costs, for the record: parsers are now stateful, which is what forced `detectSlicer` to stop handing out shared singletons, and completing an input now needs an explicit `finish()` — an API obligation the replay version did not have.

Screenshots re-captured against this build; the numbers are unchanged.

sophiedeziel:
Tidied the `finish()` call sites after review feedback.

`processGCode` and the string branch of `processGCodeStream` were parsing, finishing and executing in the same four lines twice over; that is now a shared private `ingest()`. `parser.finish()` appears twice instead of three times, in the two situations that genuinely differ — input that arrived complete, and a stream that has ended.

I did not add a wrapper around the single remaining statement. It would be one line of indirection for its own sake now that #476 no longer pairs it with anything.

The rationale for why this is not folded into `parseGCode` moved onto `Parser.finish()`, where someone reading the method will find it: making a parse assume it holds the whole input would put the obligation on every streamed chunk instead, and forgetting *that* silently brings #445 back, while forgetting this call costs only a trailing layer with no geometry.

No behaviour change — 1043 tests, 100% coverage, typecheck and lint clean. #474 and #476 rebased onto the new tip and re-verified (1044 and 1050).

PR #474, opened 2026-09-07
Stop counting a line for a chunk that completed none

sophiedeziel:
Stacked on #472. Independent of #476.

## The bug

`splitChunk` returns an empty `complete` for a chunk that finishes no line — one holding no newline, or (since #472) one holding nothing but comments, which it carries on to the next chunk. `readStream` parsed that empty string anyway, and `''.split('\n')` is `['']`, so each of those arrivals counted a line the file does not have and shifted every `lineIndex` reported after it.

That is the common case, not a rare one. A stream delivers a chunk whenever bytes arrive, not whenever a line ends, so with small chunks most arrivals complete nothing. At the harness's one-byte chunk size the parser was called once per byte, almost always with nothing to read, and a layer marker 50 characters into the file reported `lineIndex: 50`.

## The fix

```ts
if (split.complete === '') {
  continue;
}
```

Fixes the accounting, and drops a wasted parse and interpreter dispatch per byte.

The `ingestion-equivalence` fixture added in #472 had to leave `lineIndex` unasserted because of this. It now pins the exact values and `parser.lineCount` across all four ingestion modes.

## Test change worth flagging

`should flush and parse the tail when the last chunk has no trailing newline` asserted `parseGCode` was called **3** times, the middle one being the parse of `''`. Its own comment spelled that out. It was pinning the behaviour this removes, so it now expects 2.

949 tests pass, coverage stays at 100% with no new pins, typecheck and lint clean.

### DIFF
--- a/src/__tests__/gcode-preview.ts
+++ b/src/__tests__/gcode-preview.ts
@@ -561,12 +561,12 @@ describe('GCodePreview', () => {
 
       await preview.readStream(stream);
 
-      // chunk 1 completes 'G0 X0 Y0', chunk 2 has no newline so it completes
-      // nothing (''), and the leftover tail is flushed after the stream ends
-      expect(preview.parser.parseGCode).toHaveBeenCalledTimes(3);
+      // chunk 2 has no newline, so it completes nothing and is skipped rather
+      // than parsed as an empty line; the tail is flushed after the stream ends
+      expect(preview.parser.parseGCode).toHaveBeenCalledTimes(2);
       expect(preview.parser.parseGCode).toHaveBeenNthCalledWith(1, 'G0 X0 Y0');
       expect(preview.parser.parseGCode).toHaveBeenLastCalledWith('G1 X10 Y10');
-      expect(mockInterpreter.execute).toHaveBeenCalledTimes(3);
+      expect(mockInterpreter.execute).toHaveBeenCalledTimes(2);
     });
   });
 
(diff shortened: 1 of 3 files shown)


sophiedeziel:
You were right that this looked off. It was carrying a second, unrelated change — forwarding the tail flush's metadata to the job — and that change is **dead code**.

`Parser.metadata` is a single object mutated in place, and the indexer holds its `layerMetadata` array by reference, so by the time the tail runs the job is already pointing at the live data. Assigning it again cannot change anything. I tried to construct a case where it could: for no earlier chunk to have assigned metadata, every line before the last must be a comment, and that allows at most one layer block — so the layer structure is the same either way. Every variant I built came out identical with and without it.

It only looked load-bearing because the test I wrote for it handed the mock a **distinct metadata object per call**, which the real parser never does. That is the same vacuous-test trap I caught twice earlier in this stack and walked straight into here.

Both the change and its test are gone. What is left is the one thing the title claims: skipping the parse of an empty string. I also corrected the comment on it — since #472, a chunk can complete nothing because it is all comments, not only because it lacks a newline.

Diff is down to 19 insertions across three files.

sophiedeziel:
Trimmed further: **19 insertions → 14**, and the source change is now one statement plus three lines explaining why.

```ts
// Parsing '' would still count a line -- ''.split('\n') is [''] -- and
// shift every lineIndex after it. Chunks complete nothing often: bytes
// arrive without a newline, or splitChunk holds a comment run back.
if (split.complete === '') continue;
```

The rest is two forced test updates: the tail-flush test asserted the parse of `''` (3 calls, now 2), and the equivalence fixture can now pin exact `lineIndex` values and `lineCount` where it previously had to skip them.

I kept the comment rather than shrinking it to nothing — that `''.split('\n')` returns `['']` is the whole reason this line exists, and it is not obvious from reading the code.

PR #476, opened 2026-09-07
Wait for a recognised slicer to state its first layer

sophiedeziel:
Stacked on #472. Independent of #474.

## The bug

`LayersMetadataIndexer` chose between slicer metadata and tolerance detection on the **first path** it was given, and never revisited it:

```ts
this.useMetadata ??= this.hasMetadata;
```

A slicer states its layers in comments that come *after* its start G-code. So a streamed parse reaches that first path — homing, a prime line — before any layer has been read, and answers "no metadata" for a file that simply has not got to its first `;LAYER_CHANGE` yet. Metadata indexing was then off for the whole file.

A single-shot parse has the entire file in hand before any path is sorted, so it answers the opposite on the same bytes. Which answer a stream got depended on where the chunk boundaries happened to fall:

| file | 256 B | 512 B | 1 KB | 4 KB | 64 KB |
|---|---|---|---|---|---|
| calicat | fallback | fallback | metadata | metadata | metadata |
| 3DBenchy | fallback | fallback | metadata | fallback | metadata |

## The fix

Keep asking until the answer is yes.

```ts
if (!this.useMetadata && this.hasMetadata) {
  this.useMetadata = true;
  this.adoptMetadataLayers();
}
```

The lock is still real once metadata is in use — the two strategies genuinely disagree about what a layer position means — but there is nothing to protect while the answer is still "not yet".

The fallback describes the one layer it creates in the meantime, the prime line's, by the move that made it rather than by the slicer. `adoptMetadataLayers` replaces those values when metadata takes over. Measured across all five bundled files at every chunk size, the fallback creates **at most one** layer before the switch — a start G-code prints at a single height — so it is always the layer the first metadata entry describes.

## Verification

Across the five bundled demo files at 256 B, 512 B, 1 KB, 4 KB and 64 KB, every file now selects the same strategy and produces the same layer count, Z values, path counts, per-layer contents and zero duplicate filings as a single-shot parse. Before this change the small-chunk columns disagreed on strategy, and without `adoptMetadataLayers` three files still reported layer 0 at 0.28 instead of 0.3.

The equivalence-harness test uses layers rising by 0.04 — under the 0.05 tolerance — so the two strategies genuinely disagree and the test can fail; it fails on both small-chunk modes without this change.

1050 tests pass, coverage stays at 100% with no new pins, typecheck and lint clean.

## Before / after

`demo/gcodes/vase.gcode` streamed in 512 B chunks — what a slow connection or a serial feed delivers. Same file, same chunk size, same camera, and in both the layer slider is set to **1 – 220**, as a user inspecting the print would.

**Before** — the first path arrives before the file's first `;LAYER_CHANGE`, so indexing settles on the tolerance fallback. This is a vase-mode print, which the fallback rejects as non-planar, so it clears the layer index: **0 layers**. `endLayer` then clamps to nothing and the whole model keeps rendering — the slider silently does nothing.

before

**After** — the strategy waits for the metadata that was moments away, finds 332 layers, and the slider cuts the print where it is told.

after

vase.gcode is the sharpest case because the fallback cannot index it at all, but the underlying divergence is not vase-specific: at 256 B and 512 B every bundled file picked a different strategy than a single-shot parse of the same bytes.


sophiedeziel:
Tightened the end-of-input handling after review feedback.

`parser.finish()` and `job.finishIndexing()` were being called as a pair in three separate places — a two-part protocol where forgetting half of it silently loses data. They are now one private `endOfInput()`, and the whole-input path is a private `ingest()` that both `processGCode` and the string branch of `processGCodeStream` share. `parser.finish()` now appears exactly once in the codebase, and `readStream` has a single `endOfInput()` at the point where the stream is known to be over.

I did look at removing the call entirely by having `parseGCode` assume it holds the whole input and making streamed chunks opt out with a flag. Rejected it: forgetting that opt-out silently reintroduces #445, whereas forgetting an explicit end-of-input call only costs a trailing layer that has no geometry. The obligation belongs on the side where getting it wrong is cheapest, so it stays explicit — just no longer scattered.

Reordering `finish()` to after `executeCommands` in the one-shot path is safe: a block the input ends on is comments only, so no path is waiting on the layer it describes. Verified across the five bundled files at 256 B to 64 KB — same layers, path counts, per-layer contents, no duplicate filings.

1053 tests, 100% coverage, typecheck and lint clean.

sophiedeziel:
Rewrote this much smaller after review feedback — the diff went from ~200 lines across three files to 22 in one.

The previous version held paths back until the strategy was decidable, which needed a pending buffer, a release hook on the indexer, an end-of-input call on the job, and a fix to `resumeLastPath`. Measuring it against the simple alternative showed most of that was paying for itself: **the buffering was what made `resumeLastPath` double-index paths in the first place.** Without it there are no duplicates to fix.

Just re-asking the question — `??=` to a re-check — gets the correct strategy, layer count, path counts and zero duplicates at every chunk size on every bundled file. The only thing it missed was layer 0's Z (0.28 instead of 0.3 on three files), because the fallback had already created that layer from the prime line. Since the fallback provably creates at most one layer before the switch, restating it from the metadata closes that gap in eight lines.

Same end result, verified the same way, with no new public API on `Job` or the indexer.

PR #500, opened 2026-09-10
Fix inch-unit conversion for coordinates and extrusion

### LINKED ISSUE #497: Inch-based G-code is drawn 25.4 times too small
## The problem, in simple terms

Imagine reading “1 inch” on a ruler as “1 millimetre”. That is what the preview did with inch-based G-code: it drew the object 25.4 times too small and reported too little filament.

G-code tells us which ruler to use: `G20` means inches and `G21` means millimetres. For example, `G1 X10 E1` in inch mode means X = 254 mm and an extruder coordinate of 25.4 mm, not 10 mm and 1 mm. Incorrect filament lengths can also affect calculated material volume and mass.

The fix is to turn inch measurements into millimetres as each command is executed. Everything stored and drawn then uses the same ruler.

## Current status

Reviewed against [PR #500](https://github.com/xyz-tools/gcode-preview/pull/500), commit `2ad6ae1`, on 11 September 2026.

**The requested conversion and all test scenarios below are implemented. The PR is still open and unmerged.** The previous revision was approved and passed CI; the added tests pass locally, and CI for the new commit may still be running. Checked boxes mean the implementation or automated coverage is present in the PR; they do not mean the fix has shipped.

## What the fix covers

- [x] Straight moves (`G0/G1`): convert printhead coordinates and filament values.
- [x] Curved moves (`G2/G3`): convert endpoints, curve offsets/radius (`I/J/R`), and filament values.
- [x] Set coordinates (`G92`): convert printhead and extruder coordinates before applying offsets.
- [x] Probe moves (`G31/G38.x`): convert target coordinates and preserve the existing stop-at-Z0 behaviour.
- [x] Switching units changes how later commands are read; it does not rescale stored positions. Leaving out a coordinate preserves its existing value.
- [x] Curves are split into drawing segments using millimetres throughout, so equivalent inch and millimetre curves have the same smoothness.
- [x] Keep slicer comments such as `;WIDTH:` and `;HEIGHT:` in millimetres. Homing (`G28`) still needs no conversion because it goes to zero.

## Automated test coverage

The new tests run G-code through the parser and interpreter in `src/__tests__/interpreter/commands/set-units.ts`.

- [x] Inch-based straight moves and `G92` produce the expected stored coordinates in both extrusion modes (`M82/M83`).
- [x] Compare equivalent inch and millimetre **straight moves** for identical path geometry, path types, bounding dimensions, and filament length in both `M82/M83`.
- [x] Switch units partway through a file, including omitted coordinates, and check the resulting position and filament length.
- [x] Exercise `G92` in inch mode and between unit switches.
- [x] Exercise an extruder-only retraction in both `M82` and `M83`, then check the resulting extruder position after subsequent commands.
- [x] Test a complete extruder-only **retract-and-prime pair** in inch mode for both `M82/M83`. Check E after each step, unchanged printhead position/geometry/bounds, and correct filament length and geometry on the next move without a `G92` reset.
- [x] Compare equivalent inch and millimetre curves using both `I/J` and `R`, including Z movement, identical vertices/segment counts and bounding boxes, and the expected filament length.
- [x] Check every supported probe command, including Z0 contact and a previously unknown Z position.
- [x] Check that slicer width/height comments remain unchanged in inch mode.

After adding the four parameterized regression cases, local validation passes: **992 tests, 100% source coverage, type checking, lint, and build**. All four new cases fail when inch extrusion conversion is temporarily removed, confirming they catch that regression; the implementation was restored afterward. The build retains its existing `lil-gui` global-name warning. Browser verification remains with the maintainer.

## Insights from the review

- The conversion happens **before** coordinate offsets and changes in filament length are calculated. This is what makes `G92` and unit switches work together.
- The shared `toMillimeters` helper lives in `src/units.ts`, following review feedback. Each command handler still has to remember to call it. The approving reviewer suggested considering safeguards for future commands; that is a follow-up, not a blocker for this fix. A reusable inch-versus-millimetre equivalence test would help catch omissions.
- This PR covers distances and geometry. It does not establish that every unit-dependent value, such as feed rate (`F`), is normalized; avoid describing it as converting all G-code values.
- An automatically fitted camera can hide the size error. This remains relevant to #496: correct-looking framing is not proof of correct measurements.

## Remaining work

- [x] Add coverage for the two previously missing scenarios (commit `2ad6ae1`).
- [ ] Merge PR #500. Keep this issue open until the fix is merged.

No public API change is intended; the removed units argument belongs to the internal arc tessellator.

remcoder:
## Problem
Fixes #497. Inch-based G-code rendered 25.4 times too small and reported incorrect filament lengths.

## Cause and fix
G20 changed the units flag but handlers used command values directly. Distance words now become millimeters before coordinate shifts or extrusion deltas are calculated: `distance_mm = distance_in × 25.4`. Omitted words preserve existing state; G20/G21 only change interpretation of subsequent commands. G92 uses the same conversion, so its logical coordinates and stored physical offsets agree.

Arc endpoints, I/J offsets and R are normalized together. The tessellator now receives millimeters exclusively, so its chord tolerance uses the physical radius directly. Probe targets use the same conversion and retain Z0 contact behavior. Slicer width/height comments remain millimeters.

## Behavior changes
Inch files now produce correctly sized paths, bounding boxes and extrusion lengths. Millimeter files retain their existing behavior. State position documentation specifies millimeters; the internal ArcTessellator no longer accepts a units argument.

## Checks
- 992 tests passed; 100% statement, branch, function and line coverage.
- Typecheck, lint and production build passed.
- Pipeline regressions cover M82/M83, E-only moves, G92, unit switches, I/J and R helical arcs (including equal geometry and segment counts), probes, and slicer dimensions.
- Straight-move equivalence now checks geometry, path types, bounds, and filament length in both M82/M83. Inch E-only retract/prime pairs check each intermediate E position, unchanged geometry, and correct subsequent extrusion without a G92 reset.
- All four new regression cases fail when inch extrusion conversion is temporarily removed and pass with the fix restored.
- Browser verification left to the maintainer as requested.


src/state.ts
```diff
@@ -72,6 +72,11 @@ export class State {
    */
   isHomed = false;
 
+  /** Converts a command distance to millimeters, preserving omitted words. */
+  toMillimeters(value: number | undefined): number | undefined {
+    return value === undefined ? undefined : value * (this.units === 'in' ? MM_PER_INCH : 1);
+  }
```
sophiedeziel: Should this really be defined in State? That seems more like a `utils` or `units` responsibility

sophiedeziel (approved):
I like that everything is normalized early. 

I worry that bugs are easily introduced by forgetting to normalize units whenever a new command is introduced, but that's not in scope of this PR. 

Let's think a little bit about the safeguards to add.

PR #501, opened 2026-09-10
Fix filament consumption across retraction and E-only moves

### LINKED ISSUE #499: Fix extrusionDistance: include E-only extrusion and exclude retraction recovery
`JobStats.extrusionDistance` cannot currently be used as total filament consumption: it misses extrusion without XYZ movement and counts retraction recovery during XYZ movement as newly consumed filament. This affects the filament-mass analyzer motivating #496.

### 1. E-only extrusion is ignored

The no-XYZ branch in `src/interpreter/commands/linear-move.ts` updates the extruder position with `state.applyExtrusion(e)` but returns without updating `extrusionDistance`.

```gcode
M83
G1 E10
```

Actual: `extrusionDistance === 0`.
Expected: `extrusionDistance === 10` (10 mm of filament advanced without a preceding retraction).

### 2. Retraction recovery during XYZ movement is counted as new consumption

The linear and arc handlers add every positive extrusion delta to `extrusionDistance`, without accounting for earlier retractions. Filament pulled back and subsequently restored is counted again when recovery happens during movement.

```gcode
M83
G1 X0 Y0 Z0
G1 X10 E10
G1 E-1
G1 X20 E1
```

Actual: `extrusionDistance === 11`.
Expected: `extrusionDistance === 10`: the final move only restores the preceding 1 mm retraction.

## Expected semantics

Define and document `extrusionDistance` as cumulative newly advanced filament length, excluding recovery of previously retracted filament. Track outstanding retraction: negative deltas increase it; positive deltas first repay it, and only the remainder increases consumption. A retraction should not subtract from filament already consumed. This accounting must apply regardless of whether XYZ moves.

Keep path classification separate: excluding recovery from consumption must not inadvertently change rendering behavior by replacing the signed E delta used by path handlers.

## Acceptance criteria

- [x] E-only purges/primes increase consumption when there is no outstanding retraction.
- [x] Retraction followed by equal recovery adds no consumption, whether recovery is E-only, linear, or an arc.
- [x] Partial recovery leaves the remaining retraction outstanding; recovery beyond that amount counts only the excess (e.g. retract 2, advance 3 adds 1).
- [x] Absolute (`M82`) and relative (`M83`) E modes produce equivalent totals.
- [x] `G92 E` changes the coordinate reference without adding consumption or clearing outstanding retraction.
- [x] Results match for one-shot and chunked execution with the same job.
- [x] Add regression tests for both examples and the cases above, and update `JobStats.extrusionDistance` documentation.

Unit normalization (#497) is a separate issue; these examples use the default millimeter units.

remcoder:
## Problem
Fixes #499. `M83; G1 E10` reported zero filament, while retracting and restoring filament during movement counted that filament twice.

## Cause and fix
Linear and arc handlers counted only positive E deltas attached to XYZ moves. Both now send every signed delta to shared statistics accounting, including E-only moves. Outstanding retraction is tracked per stats object without expanding its public shape.

For a delta d and outstanding retraction r, newly consumed filament is `max(0, d - r)` and the remaining retraction is `max(0, r - d)`. Negative deltas therefore increase retraction without subtracting prior consumption; positive deltas repay it before adding consumption. G92 only changes coordinates and leaves this accounting intact.

## Behavior changes
`extrusionDistance` now means newly advanced filament, including stationary purges/primes and excluding recovery. The original signed delta still classifies paths, so recovery moves retain their rendering behavior. Existing command counters are unchanged.

## Checks
- 995 tests pass with 100% statement, branch, function and line coverage.
- Typecheck, lint and production build pass.
- Pipeline tests cover both issue examples, M82/M83, equal/partial/excess recovery, G92, linear and arc retractions/recovery, independent jobs, and every command chunk size in a representative sequence.
- Independent of the inch-unit fix in #500; both PRs target develop.


src/interpreter/commands/arc-move.ts
```diff
@@ -38,9 +39,7 @@ export const makeArcMove = (options: ArcTessellatorOptions = {}): CommandHandler
     const pathType = extruded > 0 ? PathType.Extrusion : PathType.Travel;
     const currentPath = job.continuePath(pathType);
 
-    if (extruded > 0) {
-      job.stats.extrusionDistance += extruded;
-    }
+    recordExtrusion(job.stats, extruded);
```
sophiedeziel: That's a weird pattern, I would expect to see an API that's more along the lines of this:
```suggestion
    job.stats.recordExtrusion(extruded);
```

src/interpreter/commands/linear-move.ts
```diff
@@ -29,7 +30,7 @@ export const linearMove: CommandHandler = (command, job) => {
     // still account the E parameter: in absolute mode a retract/prime pair
     // moves the extruder position, and losing it here would misattribute the
     // difference to the next extruding move
-    state.applyExtrusion(e);
+    recordExtrusion(job.stats, state.applyExtrusion(e));
```
sophiedeziel: Similar to the last comment, this is probably too compact and should probably be two distinct steps:


```suggestion
    extruded = state.applyExtrusion(e);
    job.stats.recordExtrusion(extruded);
```

src/__tests__/filament-consumption.ts
sophiedeziel, line 1: I noticed that coding agents INSISTS on creating a bunch of new test files that I think should be integrated into the normal class test files. 

Two observations: 
1. We don't have a clear distinction of integration tests and feature tests. 
2. There should be guidelines on how to create tests and clean up the patterns AI is probably picking up and reproducing.

sophiedeziel (approved):

PR #508, opened 2026-09-10
Add source line indices to GCodeCommand

### LINKED ISSUE #494: Add lineIndex to GCodeCommand
A parsed command knows its text, but not which line of the file it came from. Add an optional `lineIndex` so tools can point back to the right source line—even when the file is read in chunks or the same command appears more than once.

For example, a command from the fifth line would have `lineIndex: 4` (counting starts at zero).

## Why this helps

An inspector could show “this command came from line 5” without keeping its own counter.

It also helps match slicer settings, such as extrusion width, to the right commands. Today the job counts commands to work out the source line. Normal preview loading works, but a custom pipeline that removes comment-only commands after reading their settings can throw that count off. This was reproduced; it is not a known bug in normal preview loading.

## Backwards-compatible approach for 3.1+

- [ ] Add `lineIndex?: number` to `GCodeCommand`. Keep its constructor unchanged.
- [ ] Set it in `parseGCode`, counting source lines across chunks, including blank lines and comments.
- [ ] Leave it undefined for manually created commands and standalone `parseCommand` calls, where the file position is unknown.
- [ ] Pass the index to `Job.beginCommand` as an optional argument. When no index is supplied, keep the existing counter as a fallback so old callers still work.
- [ ] Keep `parseCommand`'s existing signature and return type. Any cleanup can be separate.
- [ ] Test chunked input, repeated lines, blank/comment lines, old callers without indices, and metadata matching after comment-only commands are removed.

Keep the empty-chunk guard in `readStream`: an empty chunk must not invent an extra source line. Commands must still execute in their original order; a line number does not make reordering G-code safe.

This can ship after 3.0 without requiring callers to change their code. It does not need to block the standalone parser entry point in #492. The compatibility tradeoff is keeping the fallback counter.

remcoder:
## Summary

Fixes #494.

Commands now carry their zero-based source line index, making the original location available to consumers and allowing extrusion metadata to be selected without counting executed commands. Indices span successive `parseGCode` calls, including when source lines are not retained.

The normal execution path still advances a metadata cursor. A command from an earlier source line rewinds and replays dimension events, preventing later widths or heights from leaking backwards. Empty stream chunks are passed as zero lines, preserving source coordinates without the old early-continue guard.

## Behavior changes

Filtering comment-only commands preserves the extrusion widths and heights associated with the remaining commands. Reordering commands selects dimensions from each command's original source position; motion and other interpreter state still follow execution order. Ordinary file and stream ingestion retain their existing geometry, covered by the ingestion equivalence suite.

## Public API changes

This is a breaking change for 3.0:

- `GCodeCommand` takes `(lineIndex, src, gcode, params, comment?)` and exposes `lineIndex: number`.
- `Parser.parseCommand(line, keepComments?, lineIndex?)` accepts an optional source index, defaulting to zero for standalone calls, and returns `GCodeCommand` rather than `GCodeCommand | null`. Standalone calls do not advance the parser's line count.
- `Job.beginCommand(command)` now requires the command whose metadata should be applied.

## Checks

- [x] 984 tests pass, including compile-time public API assertions; all source coverage is 100%.
- [x] `npm run typeCheck`
- [x] `npm run lint`
- [x] `npm run build`
- [x] Confirmed both filtered/reordered regression tests fail when the former execution-count behavior is restored.

Tests ran with `NODE_OPTIONS=--no-experimental-webstorage` on local Node 26 so happy-dom supplies storage. Build emits the existing Rollup warning about the `lil-gui` global name. Browser verification was omitted as requested.


src/job.ts
```diff
@@ -106,18 +104,20 @@ export class Job {
...
    * the last path at every chunk boundary, undoing any break performed here.
    */
-  beginCommand(): void {
-    const lineIndex = this.executedCommandCount++;
+  beginCommand({ lineIndex }: GCodeCommand): void {
+    if (this.dimensionCursor > 0 && this.extrusionDimensions[this.dimensionCursor - 1].lineIndex > lineIndex) {
+      this.dimensionCursor = 0;
+      this.state.extrusionWidth = undefined;
+      this.state.lineHeight = undefined;
+    }
```
sophiedeziel, line 120: It took me a while to understand why that logic is there. But I validate that it is correct.

sophiedeziel (approved):

PR #509, opened 2026-09-10
Enable tube rendering by default

remcoder:
Extrusions now render as tubes when `renderTubes` is omitted. Callers can still select flat lines with `renderTubes: false`.

Updated the README and existing tests to cover the new default and explicit line mode.

Validation: 978 tests pass with 100% coverage, build, type checking, and lint pass. Tests ran with `NODE_OPTIONS=--no-experimental-webstorage` for the local Node 26 environment.

### DIFF
--- a/src/__tests__/objects-manager.ts
+++ b/src/__tests__/objects-manager.ts
@@ -14,6 +14,7 @@ describe('ObjectsManager', () => {
   beforeEach(() => {
     scene = new Scene();
     objectsManager = new ObjectsManager(scene, 0.4, 0.2, 0.6);
+    objectsManager.renderTubes = false;
   });
 
   describe('constructor', () => {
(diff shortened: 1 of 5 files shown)


src/__tests__/objects-manager.ts
```diff
@@ -14,6 +14,7 @@ describe('ObjectsManager', () => {
   beforeEach(() => {
     scene = new Scene();
     objectsManager = new ObjectsManager(scene, 0.4, 0.2, 0.6);
+    objectsManager.renderTubes = false;
```
sophiedeziel, line 17: Shouldn't the tests apply the defaults "by default"?

src/objects-manager.ts
```diff
@@ -67,7 +67,7 @@ export class ObjectsManager {
    * is unset too.
    */
   extrusionWidth?: number;
-  renderTubes = false;
+  renderTubes = true;
```
sophiedeziel, line 70: 🎉

PR #532, opened 2026-09-14
fix: clear preset state when dropping a file

### LINKED ISSUE #529: Demo: dropping a file keeps the previous preset's model credits and thumbnail
## What happens

Load any preset (e.g. 3DBenchy), then drag & drop a G-code file onto the canvas. The info widget updates the file name, size, layer count and slicer, but the **model credits and the thumbnail still belong to the previously selected preset**.

So a dropped file ends up displayed under someone else's model name, designer and license:

```
3DBenchy
by CreativeTools
CC0 - public domain
dropped-test.gcode (733.57 kB)
Sliced with PrusaSlicer
```

The attribution is wrong rather than just stale, which is the part that matters — it credits a model the user did not load.

## Why

`model` is only ever assigned in `selectPreset()` in `demo/js/app.js`, from `preset.model`. Nothing clears it on drop.

The thumbnail has the same shape of problem: `updateUI()` sets `thumbnail.value` from the parser metadata, so a dropped file whose slicer embedded no thumbnail leaves the previous preset's thumbnail on screen.

## Suggested fix

Clear the preset-derived state when a dropped file starts loading. The demo already listens for the library's `update` event on the canvas (added in #526) — that's a natural place, though clearing on `drop`/stream start would show the stale data for a shorter time than clearing when the stream ends.

Also worth checking whether `selectedPreset` should be cleared, since the dropdown keeps showing a preset that is no longer what's rendered.

## Notes

Pre-existing; noticed while verifying the file size display in #526.

EthemKD:
Closes #529

## Summary
- clear preset-derived demo state when a user drops a local G-code file
- reset the selected preset, thumbnail, and preset model attribution before the dropped file loads
- guard the preset watcher so clearing the selection does not try to re-apply a preset
- add regression coverage for clearing preset-derived state on real file drops

## Validation
- `npx.cmd vitest run src/__tests__/demo-utils.ts` — passed
- `npx.cmd vitest run` — passed
- `npx.cmd vitest run --coverage` — passed
- `npm.cmd run typeCheck` — passed
- `npm.cmd run lint` — passed
- `npm.cmd run build` — passed
- `git diff --check` — passed

## Manual verification
Before:
- selected preset: `3DBenchy`
- preset thumbnail and attribution visible

After dropping `dropped-test.gcode` onto the preview canvas:
- preset selection is cleared
- preset thumbnail disappears
- preset attribution disappears
- dropped file info is shown instead

AI-assisted during development; I reviewed, tested, and manually validated the change.

Ekran görüntüsü 2026-09-15 022706
Ekran görüntüsü 2026-09-15 022251


demo/js/utils.js
```diff
@@ -24,3 +24,11 @@ export function parseIntOrDefault(value, defaultValue = 0) {
   const parsed = parseInt(value, 10);
   return isNaN(parsed) ? defaultValue : parsed;
 }
+
+export function clearPresetStateOnFileDrop(event, state) {
```
remcoder, line 28: I think fileDropHandler is a better name for this function.

demo/js/app.js
```diff
@@ -113,6 +113,12 @@ export const app = (window.app = createApp({
     onMounted(async () => {
       const canvas = document.querySelector('canvas.preview');
 
+      canvas.addEventListener(
+        'drop',
+        (evt) => clearPresetStateOnFileDrop(evt, { selectedPreset, model, thumbnail }),
+        { capture: true }
```
remcoder, line 119: Is capture mode a requirement here? If so, please add a comment to the code to explain, If not I prefer to default to bubbling.

remcoder (changes requested):
Thanks for you contribution! Much appreciated. Please make the requested changes so this PR can be merged.

